PR #134, opened 2026-06-13
Guard against null ResultSet in EnvironmentRepositoryImpl

### LINKED ISSUE #132: EnvironmentRepositoryImpl lacks null-ResultSet guard present in sibling repositories
## Summary
`EnvironmentRepositoryImpl` query methods (`findAll`, `findById`, `findByName`, `findByEntityId`, `findEntityIdsByEnvironmentId`, `findLocationIdsByEnvironmentId`, `findGridIdsByEnvironmentId`) call `rs.next()` directly without first checking whether the `ResultSet` returned by `DbInteractions.query(...)` is `null`.

Its sibling repositories guard against this. For example `GridRepositoryImpl` has:

```java
ResultSet rs = dbInteractions.query(...);
if (rs == null) {
    return ...; // empty result
}
```

`EntityRepositoryImpl` and `LocationRepositoryImpl` do the same.

## Impact
When `query(...)` returns `null` (the signal for a failed query), the read methods on `EnvironmentRepositoryImpl` throw `NullPointerException` instead of returning an empty result, diverging from the behaviour of every other repository.

## Why this was filed (not fixed) here
Discovered while adding `EnvironmentRepositoryImplTest` (#129). That PR is test-only; per the dev-loop test-expansion rules, production behaviour is not changed under a test cycle. The new tests therefore characterize current behaviour and intentionally omit null-ResultSet cases.

## Proposed work
- Add the `if (rs == null)` guard to each query method in `EnvironmentRepositoryImpl`, matching the sibling repositories.
- Add the `...ReturnsEmptyWhenResultSetIsNull` test cases to `EnvironmentRepositoryImplTest` (mirroring the sibling tests) once the guard is in place.

---

dmccoystephenson:
## Summary
- Adds the missing `if (rs == null)` guard to the **seven** query methods of `EnvironmentRepositoryImpl` (`findAll`, `findById`, `findByName`, `findByEntityId`, `findEntityIdsByEnvironmentId`, `findLocationIdsByEnvironmentId`, `findGridIdsByEnvironmentId`), matching the existing pattern in `GridRepositoryImpl`, `EntityRepositoryImpl`, and `LocationRepositoryImpl`.
- Before this change those methods called `rs.next()` directly, so a `null` `ResultSet` — the value `DbInteractions.query(...)` returns on failure — produced a `NullPointerException` instead of an empty result, diverging from every sibling repository.
- Each guard logs with the method's existing message prefix (`"... : ResultSet is null"`) and returns the empty result (`Optional.empty()` / empty list), exactly as the siblings do.

## Tests
- Adds the seven `...ReturnsEmptyWhenResultSetIsNull` cases to `EnvironmentRepositoryImplTest`. **These are the regression guard:** without the new `rs == null` check each one throws `NullPointerException` at `rs.next()` and fails.
- Removes the now-obsolete class-level comment that documented the omission.

## Test plan
- [x] `./mvnw test -Dtest=EnvironmentRepositoryImplTest` — 45/45 pass (JDK 21)
- [x] `./mvnw test` (full suite) — 213/213 pass, no regressions
- [ ] Python client — not applicable (no `src/main/python` change)

## Notes
- The write/delegator methods (`save`, `deleteById`, `updateName`, `delete*`) are unchanged: they delegate to `DbInteractions.update`, which returns a `boolean` and has no `ResultSet` to guard.

Closes #132

---

### DIFF
--- a/src/main/java/preponderous/viron/repositories/EnvironmentRepositoryImpl.java
+++ b/src/main/java/preponderous/viron/repositories/EnvironmentRepositoryImpl.java
@@ -26,6 +26,10 @@ public class EnvironmentRepositoryImpl implements EnvironmentRepository {
     public List<Environment> findAll() {
         List<Environment> environments = new ArrayList<>();
         ResultSet rs = dbInteractions.query("SELECT * FROM viron.environment");
+        if (rs == null) {
+            log.error("Error finding all environments: ResultSet is null");
+            return environments;
+        }
         try {
             while (rs.next()) {
                 environments.add(mapResultSetToEnvironment(rs));
@@ -39,6 +43,10 @@ public class EnvironmentRepositoryImpl implements EnvironmentRepository {
     @Override
     public Optional<Environment> findById(int id) {
         ResultSet rs = dbInteractions.query("SELECT * FROM viron.environment WHERE environment_id = " + id);
+        if (rs == null) {
+            log.error("Error finding environment by id: ResultSet is null");
+            return Optional.empty();
+        }
         try {
             if (rs.next()) {
                 return Optional.of(mapResultSetToEnvironment(rs));
@@ -52,6 +60,10 @@ public class EnvironmentRepositoryImpl implements EnvironmentRepository {
     @Override
     public Optional<Environment> findByName(String name) {
         ResultSet rs = dbInteractions.query("SELECT * FROM viron.environment WHERE name = '" + name + "'");
+        if (rs == null) {
+            log.error("Error finding environment by name: ResultSet is null");
+            return Optional.empty();
+        }
         try {
             if (rs.next()) {
                 return Optional.of(mapResultSetToEnvironment(rs));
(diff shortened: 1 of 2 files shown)


dmccoystephenson (commented):
Self-review rubric (adversarial; anchored on green CI run 27466954846 + a direct regression demonstration):

- **Scope:** PASS — only two files, both required for #132: `EnvironmentRepositoryImpl.java` (the guards) and its test. No unrelated edits.
- **Tests-fix:** PASS — **demonstrated, not asserted.** Reverting just the production file to `main` and re-running the new null tests errors with `NullPointerException: Cannot invoke "ResultSet.next()" because "rs" is null` on all sampled cases; with the guard they pass. Seven `...ReturnsEmptyWhenResultSetIsNull` tests, one per guarded method.
- **Tests-new:** N/A — no new public methods; guards added to existing ones.
- **Sibling structure:** PASS — each guard mirrors `GridRepositoryImpl` exactly: `if (rs == null) { log.error("<method prefix>: ResultSet is null"); return <empty>; }`, reusing each method's existing log-message prefix.
- **Sibling renames:** N/A — no identifiers renamed.
- **Docs:** N/A — the change is internal error-handling (null → empty result); no API/DTO/endpoint contract or Phase 7 doc source-of-truth row is affected.
- **Issue resolution:** PASS — #132's named surface (the seven query methods lacking the guard) is fully addressed; `grep` confirms 7 guards + 7 null tests.
- **CI:** PASS — `build` green on PR head.
- **DTO boundary / Spec alignment:** N/A — no controller or endpoint change.
- **Java/Python parallelism:** PASS — repository layer is Java-only; the Python client has no repository mirror.
- **Override correctness:** PASS — `git diff` shows zero `@Override` lines added or changed in production.

**Observation (not a FAIL, pre-existing & repo-wide):** treating a null ResultSet (a failed query) as an empty result silently hides DB failures from callers. That is the established convention across all four repositories, so this PR conforms rather than introduces it; changing the convention is out of scope.

Summary: ready to merge — the fix is small, matches the sibling pattern, and the regression is proven by the new tests failing without the guard.

Issue #128, opened 2026-06-13
Align write endpoints with spec: use request-body DTOs instead of path variables


dmccoystephenson:
## Summary
The write endpoints accept their inputs as **path variables** rather than request bodies, which diverges from `docs/MVP.md` and `docs/openapi/viron-api.json`. No controller currently uses `@RequestBody`.

## Spec vs. implementation
| Operation | Spec (MVP.md / OpenAPI) | Implemented |
|-----------|-------------------------|-------------|
| Create entity | `POST /api/v1/entities` with body | `POST /api/v1/entities/{name}` (path var) |
| Create environment | body `CreateEnvironmentRequest` | `POST /api/v1/environments/{name}/{numGrids}/{gridSize}` (path vars) |
| Rename environment | body `UpdateEnvironmentNameRequest` | `PATCH /api/v1/environments/{id}/name/{name}` (path var) |

As a consequence, the two request DTOs named in MVP.md — **`CreateEnvironmentRequest`** and **`UpdateEnvironmentNameRequest`** — were never created (`dto/` holds only response DTOs).

## Proposed work
- Add `CreateEnvironmentRequest` and `UpdateEnvironmentNameRequest` DTOs with bean validation.
- Accept request bodies on the create/rename/create-entity endpoints (a `CreateEntityRequest` may be warranted).
- Decide whether the path-variable forms are kept for backward compatibility or removed; update `docs/openapi/viron-api.json` either way so spec and code agree.

## Notes
A decision is needed on whether the spec or the current path-variable behaviour is authoritative; this issue assumes the spec is.

---
*Filed by Claude on behalf of Daniel Stephenson.*

---

_drafted by Claude on behalf of Daniel Stephenson_

dmccoystephenson:
Deferred from the 2026-06-13 dev-loop cycle (which shipped #133, the EnvironmentRepositoryImpl tests). Skip reason: this issue requires a human decision on whether the OpenAPI spec or the current path-variable behaviour is authoritative, and it modifies the wire contract `docs/openapi/viron-api.json` — a do-not-auto-merge path. Better handled in a human-gated cycle.

dmccoystephenson:
[...]

Issue #136, opened 2026-06-13
Postman collection is wholesale stale (pre-rebuild GET-based paths)


dmccoystephenson:
## Summary
`postman/Viron.postman_collection.json` predates the Spring Boot rebuild. Every request is **GET** and uses the old non-versioned path scheme (e.g. `localhost:9999/environment/get-environment-by-id/1`, `/create-environment/{name}/{numGrids}/{gridSize}`, `/set-environment-name/{id}/{name}`) rather than the current `/api/v1/...` REST verbs and paths.

## Impact
The collection does not reflect the current API for **any** endpoint, so it is not a usable source of truth. This was noticed during #128: the create/rename environment entries there are GET-based and don't match even the pre-#128 Spring API, so #128 deliberately did not partially patch them.

## Proposed work
- Regenerate or rewrite the collection against the current API (correct base path `/api/v1`, correct verbs, request bodies for the write endpoints aligned with `docs/openapi/viron-api.json`).
- Consider generating it from the OpenAPI spec to keep it in sync (relates to #130).

---
*Filed by Claude on behalf of Daniel Stephenson.*

Issue #135, opened 2026-06-13
Align entity write endpoints with request bodies + add entity resource to OpenAPI spec


dmccoystephenson:
## Summary
Split from #128. The environment write endpoints were aligned to request bodies in PR (see #128). The **entity** write endpoints remain on the path-variable anti-pattern, and — unlike the environment endpoints — they are **not defined in `docs/openapi/viron-api.json` at all**, so aligning them also requires authoring the entity resource in the spec.

## Current state (path variables)
- `POST /api/v1/entities/{name}` — should be `POST /api/v1/entities` with a `CreateEntityRequest` body (`{name}`), per `docs/MVP.md`.
- `PATCH /api/v1/entities/{id}/name/{name}` — should be `PATCH /api/v1/entities/{id}/name` with an `UpdateEntityNameRequest` body (`{name}`).

## Proposed work
- Add `CreateEntityRequest` and `UpdateEntityNameRequest` DTOs (with `@NotBlank` validation), mirroring `CreateEnvironmentRequest` / `UpdateEnvironmentNameRequest`.
- Switch the two `EntityController` write methods to `@Valid @RequestBody`.
- Update `EntityController` MockMvc tests and the Python client (`entityService`) + its tests.
- **Add the entity resource to `docs/openapi/viron-api.json`** (paths + request/response schemas) so the spec covers entities. This touches the wire contract, so the resulting PR will need human review (do-not-auto-merge).

## Related spec discrepancy noticed while doing #128
- The spec says `POST /api/v1/environments` returns **200**, but the controller returns **201 Created** (`@ResponseStatus(HttpStatus.CREATED)`). 201 is the more correct REST status for creation; the spec should likely be updated to 201. Tracked here for a spec-accuracy pass (not changed under the #128 environment PR, which was input-shape-only).

---
*Filed by Claude on behalf of Daniel Stephenson.*

PR #137, opened 2026-06-13
Use request-body DTOs for environment create/rename endpoints

dmccoystephenson:
## Summary
Aligns the two **environment** write endpoints with the OpenAPI contract, which already specifies request bodies and the `CreateEnvironmentRequest` / `UpdateEnvironmentNameRequest` schemas that were never implemented (the gap #128 calls out):

| Endpoint | Before | After |
|---|---|---|
| Create | `POST /api/v1/environments/{name}/{numGrids}/{gridSize}` | `POST /api/v1/environments` + `CreateEnvironmentRequest` body `{name, numGrids, gridSize}` |
| Rename | `PATCH /api/v1/environments/{id}/name/{name}` | `PATCH /api/v1/environments/{id}/name` + `UpdateEnvironmentNameRequest` body `{name}` |

- New DTOs carry the same validation the path variables enforced (`@NotBlank` name, `@Min(1)` numGrids/gridSize) via `@Valid @RequestBody`; invalid input returns **400** through the existing `GlobalExceptionHandler`.
- Keeps **all three representations** in lockstep:
  - **Server** controller (`EnvironmentController`).
  - **Java client SDK** (`services/EnvironmentService`, consumed by `DebugController`) — now sends `HttpEntity` bodies to the new paths instead of the old path-variable URLs (would otherwise 404).
  - **Python client** (`services/environmentService`).
- Updates the **MockMvc tests** (rewrote the 6 create/rename tests for JSON bodies; added 3 validation tests: blank name, non-positive numGrids, blank rename) and the **Python client tests**.
- Updates the drifted **`docs/MVP.md` / `docs/PLANNING.md`** endpoint descriptions, which still documented the path-variable forms.
- **No change to `docs/openapi/viron-api.json`** — the spec was already correct; the code is now aligned to it.

## Scope / split
This resolves the **environment** half of #128. The **entity** write endpoints are the same anti-pattern but are **not defined in the OpenAPI spec at all**, so aligning them additionally requires authoring the entity resource in `viron-api.json` (a wire-contract change needing human review). That work is split into **#135** to keep this PR coherent and auto-mergeable. Pre-existing wholesale Postman collection drift is split into **#136**.

Partially addresses #128 (environment endpoints); does not auto-close it — see #135 for the remaining entity portion.

## Known out-of-scope discrepancy
The spec says `POST /api/v1/environments` returns **200**, but the controller returns **201 Created** (more correct for a creation). This PR is input-shape-only and leaves the status as-is; tracked in #135 for a spec-accuracy pass.

## ⚠️ Breaking change
The two endpoints' URLs change. Acceptable pre-MVP; the spec was always the intended contract (confirmed with the maintainer).

## Test plan
- [x] `./mvnw test` (full suite) — **216/216 pass** (was 213; +3 validation tests), JDK 21
- [x] `pytest src/test/python` — **80/80 pass** (Python 3.11). CI does **not** run Python, so this local run is the Python anchor.
- [x] Endpoints verified against `docs/openapi/viron-api.json` (path, verb, body schema match exactly)
- [x] Java client SDK + Python client both updated to the new contract

---

### DIFF
--- a/docs/MVP.md
+++ b/docs/MVP.md
@@ -23,8 +23,8 @@ Deliver a working, tested API that supports:
 - [ ] `GET /api/v1/environments/{id}` – Retrieve a specific environment by ID.
 - [ ] `GET /api/v1/environments/name/{name}` – Retrieve a specific environment by name.
 - [ ] `GET /api/v1/environments/entity/{entityId}` – Get the environment containing a specific entity.
-- [ ] `POST /api/v1/environments/{name}/{numGrids}/{gridSize}` – Create an environment.
-- [ ] `PATCH /api/v1/environments/{id}/name/{name}` – Update environment name.
+- [ ] `POST /api/v1/environments` – Create an environment (JSON body: `name`, `numGrids`, `gridSize`).
+- [ ] `PATCH /api/v1/environments/{id}/name` – Update environment name (JSON body: `name`).
 - [ ] `DELETE /api/v1/environments/{id}` – Delete an environment and all related entities, locations, and grids.
 
 ---
--- a/docs/PLANNING.md
+++ b/docs/PLANNING.md
@@ -62,9 +62,9 @@ Deliver GET endpoints for all resources.
 ### **M4 – Write APIs**
 Implement creation, update, and deletion endpoints.
 
-19. **POST /environments/{name}/{numGrids}/{gridSize}** — Create environment with grids and locations.  
+19. **POST /environments** — Create environment with grids and locations (JSON body: `name`, `numGrids`, `gridSize`).  
 20. **DELETE /environments/{id}** — Delete environment with related data.  
-21. **PATCH /environments/{id}/name/{name}** — Update environment name.  
+21. **PATCH /environments/{id}/name** — Update environment name (JSON body: `name`).  
 31. **PUT /locations/{locationId}/entity/{entityId}** — Add entity to location.  
 32. **DELETE /locations/{locationId}/entity/{entityId}** — Remove entity from a specific location.  
 33. **DELETE /locations/entity/{entityId}** — Remove entity from its current location.  
--- a/src/main/java/preponderous/viron/controllers/EnvironmentController.java
+++ b/src/main/java/preponderous/viron/controllers/EnvironmentController.java
@@ -9,7 +9,9 @@ import lombok.extern.slf4j.Slf4j;
 import org.springframework.http.HttpStatus;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.*;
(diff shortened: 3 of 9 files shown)


dmccoystephenson (commented):
Self-review rubric (adversarial; anchored on green CI run 27467410219 + the diff):

- **Scope:** PASS — 9 files, all required for the environment write-endpoint refactor (2 DTOs, controller, Java client SDK, Python client, both test surfaces, the two drifted docs). No unrelated churn.
- **Tests-new:** PASS — new endpoint behaviour exercised by the rewritten 6 MockMvc create/rename tests + 3 new validation tests (blank name, non-positive numGrids, blank rename); Python client tests updated to assert the JSON bodies.
- **Sibling structure:** PASS — `CreateEnvironmentRequest` / `UpdateEnvironmentNameRequest` mirror `EnvironmentDto` (Lombok `@Data`/`@NoArgsConstructor`/`@AllArgsConstructor` + `@Schema`) and form a parallel request-DTO pair.
- **Sibling renames:** PASS — **the key item.** The endpoint-shape change was applied across the whole parallel set: server controller, **Java client SDK** (`services/EnvironmentService`), Python client, both test surfaces, and `MVP.md` + `PLANNING.md`.
- **Docs:** PASS — `MVP.md` + `PLANNING.md` updated; `README` already used the body form; `viron-api.json` already correct (no change). Pre-existing wholesale Postman drift split to #136 rather than partially patched.
- **Issue resolution:** PASS (honest partial) — resolves the environment half of #128; the entity half is explicitly split to #135 and this PR does **not** auto-close #128. No false closure.
- **CI:** PASS — `build` green on PR head.
- **DTO boundary:** PASS — `createEnvironment` still returns `EnvironmentDto`; no internal model (`Environment`) leaked.
- **Spec alignment:** PASS — `POST /api/v1/environments` (body `CreateEnvironmentRequest`) and `PATCH /api/v1/environments/{id}/name` (body `UpdateEnvironmentNameRequest`) match `viron-api.json` exactly. The pre-existing `200`-vs-`201` response-status discrepancy is left as-is and tracked in #135.
- **Java/Python parallelism:** PASS — server + Java client + Python client all updated.
- **Override correctness:** PASS — `git diff` shows zero `@Override` churn in production.

**Substantive finding surfaced AND fixed during review:** the first push updated the server, the Python client, and the docs but **missed the Java client SDK** (`services/EnvironmentService`, consumed by `DebugController`), which still built the old path-variable URLs and would have 404'd at runtime. CI would **not** have caught it (no test covers that client). The adversarial parallelism grep caught it; fixed in the second commit (88129d0). This is exactly the cross-representation drift this repo's loop exists to prevent.

Summary: ready to merge as the environment half of #128; entity half tracked in #135, Postman drift in #136.

PR #143, opened 2026-06-13
Parameterize SQL and close JDBC resources in DbInteractions (#139, #140)

### LINKED ISSUE #139: Data layer builds SQL by string-concatenating user-supplied names (injection / breakage risk)
## Summary
Across the factories and repositories, SQL is assembled by concatenating (or `String.format`-ing) caller-supplied values — including user-controlled entity/environment **names** — directly into the query string, with no parameterization or escaping. A name containing a single quote breaks the query, and a crafted name can alter it (SQL injection).

Confirmed occurrences (user-controlled `name` concatenated):
- `src/main/java/preponderous/viron/factories/EntityFactory.java:30`
  ```java
  String query = "INSERT INTO viron.entity (entity_id, name, creation_date) VALUES (" + id + ", '" + name + "', '" + creationDate + "')";
  ```
- `src/main/java/preponderous/viron/repositories/EntityRepositoryImpl.java:155`
  ```java
  String query = String.format("INSERT INTO viron.entity (name, creation_date) VALUES ('%s', NOW())", entity.getName());
  ```
- `src/main/java/preponderous/viron/repositories/EntityRepositoryImpl.java:180`
  ```java
  String query = String.format("UPDATE viron.entity SET name = '%s' WHERE entity_id = %d", name, id);
  ```
- `src/main/java/preponderous/viron/repositories/EnvironmentRepositoryImpl.java:62`
  ```java
  ResultSet rs = dbInteractions.query("SELECT * FROM viron.environment WHERE name = '" + name + "'");
  ```
- `src/main/java/preponderous/viron/repositories/EnvironmentRepositoryImpl.java:107`
  ```java
  String query = "UPDATE viron.environment SET name = '" + name + "' WHERE environment_id = " + id;
  ```

`EnvironmentFactory` builds its INSERTs the same way (`factories/EnvironmentFactory.java:37`, `:56`, …). These are representative, not exhaustive — the pattern is systemic to the data layer.

## Why it matters
- **Correctness:** any name containing `'` (e.g. `O'Brien`) produces malformed SQL and the operation fails.
- **Security:** a name such as `'); DROP TABLE viron.entity; --` is interpolated into an executable statement — a classic injection vector reachable from the create/rename endpoints.

## Suggested fix
Move the data layer to **parameterized queries** (`PreparedStatement` with `?` placeholders, or named parameters via Spring's `JdbcTemplate`/`NamedParameterJdbcTemplate`). This requires extending the `DbInteractions` API (`database/DbInteractions.java`) to accept bind parameters, then updating the factories and repositories to pass values as parameters rather than concatenated literals. Doing so also removes the manual quoting throughout.

### LINKED ISSUE #140: DbInteractions.query/update never close the Statement (JDBC resource leak)
## Summary
Both methods in `DbInteractions` create a `Statement` via `connection.createStatement()` but never hold a reference to it or close it, so every query/update leaks a `Statement` (and, for `query`, the returned `ResultSet` is likewise never closed by callers).

- `src/main/java/preponderous/viron/database/DbInteractions.java:32-39`
  ```java
  public ResultSet query(String query) {
      try {
          return connection.createStatement().executeQuery(query);   // line 34 — Statement is unreferenced, never closed
      } catch (SQLException e) {
          log.error("Error executing query: {}", e.getMessage());
      }
      return null;
  }
  ```
- `src/main/java/preponderous/viron/database/DbInteractions.java:41-49`
  ```java
  public boolean update(String query) {
      try {
          int rowCount = connection.createStatement().executeUpdate(query);   // line 43 — same leak
          return rowCount > 0;
      } catch (SQLException e) {
          log.error("Error executing update: {}", e.getMessage());
      }
      return false;
  }
  ```

Callers of `query(...)` (every repository) iterate the returned `ResultSet` but never close it either — and since the creating `Statement` is unreachable, neither can be closed afterward.

## Why it matters
`query`/`update` are called on essentially every request (repositories, factories). On a real Postgres connection each leaked `Statement`/`ResultSet` holds server- and driver-side resources until GC; under sustained load this risks cursor/statement exhaustion and connection instability. These leaks are invisible to the current tests because `DbInteractions` is mocked everywhere.

## Suggested fix
- For `update(...)`: wrap in try-with-resources — `try (Statement st = connection.createStatement()) { return st.executeUpdate(query) > 0; }`.
- For `query(...)`: returning a live `ResultSet` makes ownership ambiguous; prefer changing the method to accept a row-mapper / return a materialized result (e.g. a `List`) so the `Statement` and `ResultSet` can be closed inside the method. (This pairs naturally with the parameterized-query refactor in the SQL-injection issue.)

dmccoystephenson:
## Summary
Hardens the hand-rolled data layer to fix the injection/breakage risk (#139) and the JDBC resource leak (#140) together, since both live in `DbInteractions`.

**`DbInteractions` rework:**
- `query(sql, params...)` — binds parameters via `PreparedStatement`, runs inside try-with-resources, and returns a **disconnected `CachedRowSet`** (rows copied out) so the `PreparedStatement` and live `ResultSet` are always closed. Callers keep using the `ResultSet` API unchanged. Fixes the leak in #140.
- `update(sql, params...)` — binds parameters, closes the statement via try-with-resources. Fixes the leak in #140.
- Both are **varargs**, so the ~40 existing parameterless call sites compile and behave unchanged.

**Parameterization (#139):** the user-controlled `name` interpolations were moved off string concatenation to bound parameters:
- `EntityFactory` / `EnvironmentFactory` inserts
- `EntityRepositoryImpl.save` + `updateName`
- `EnvironmentRepositoryImpl.findByName` + `updateName`

Integer ids are left inline (not an injection vector); the grid/location inserts in `EnvironmentFactory` are int-only and untouched. The affected repository test mocks were updated to the parameterized call signatures.

## Tests
Adds `DbInteractionsTest` (**H2 in-memory, test scope**) — this is what actually exercises the new binding/resource behavior, since `DbInteractions` is mocked everywhere else:
- parameters bind and the row round-trips;
- **a value containing a SQL payload (`O'Brien'); DROP TABLE person; --`) is stored literally and does not inject** (the table survives) — the core #139 proof;
- the returned `ResultSet` is disconnected and fully readable after `query()` returns (evidence for the #140 resource fix);
- invalid SQL returns `null`/`false` without throwing; a no-row update returns `false`.

## ⚠️ Needs human review — not auto-merged
This PR modifies **`pom.xml`** (adds the `com.h2database:h2` test-scope dependency), which is on the do-not-auto-merge list. Flagging for manual review rather than merging automatically. The `query()` redesign (returning a `CachedRowSet` instead of a live `ResultSet`) is also a deliberate behavioral change in the core data layer worth a careful look.

## Test plan
- [x] `./mvnw test` — **226/226 pass** (was 220; +6 `DbInteractionsTest`), JDK 21
- [x] `DbInteractionsTest` runs real SQL on H2, including the injection-payload case
- [ ] Python client — n/a (no `src/main/python` change)

## Follow-up note
The deeper fix for `query()` returning a `ResultSet` at all (a `RowMapper`-style API returning `List<T>`) would let every repository drop its manual `ResultSet` handling, but that is a much larger refactor; the `CachedRowSet` approach here fixes the leak with minimal blast radius.

Closes #139
Closes #140

---

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,6 +29,13 @@
             <scope>test</scope>
         </dependency>
 
+        <!-- In-memory database used only to exercise DbInteractions (parameter binding + resource handling) in tests. -->
+        <dependency>
+            <groupId>com.h2database</groupId>
+            <artifactId>h2</artifactId>
+            <scope>test</scope>
+        </dependency>
+
         <dependency>
             <groupId>org.postgresql</groupId>
             <artifactId>postgresql</artifactId>
--- a/src/main/java/preponderous/viron/database/DbInteractions.java
+++ b/src/main/java/preponderous/viron/database/DbInteractions.java
@@ -5,8 +5,12 @@ package preponderous.viron.database;
 
 import java.sql.Connection;
 import java.sql.DriverManager;
+import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+
+import javax.sql.rowset.CachedRowSet;
+import javax.sql.rowset.RowSetProvider;
 
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -29,23 +33,57 @@ public class DbInteractions {
         this.connection = connect();
     }
 
-    public ResultSet query(String query) {
-        try {
-            return connection.createStatement().executeQuery(query);
+    /**
+     * Execute a parameterized SELECT and return a disconnected result.
+     *
+     * <p>The query is run through a {@link PreparedStatement} so caller-supplied
+     * values are bound as parameters rather than concatenated into SQL. The
+     * statement and live result set are closed before returning; the rows are
+     * copied into a {@link CachedRowSet} so callers can read them without leaking
+     * JDBC resources.
+     *
+     * @param query  SQL with {@code ?} placeholders for each parameter
+     * @param params values to bind to the placeholders, in order
+     * @return a disconnected {@link ResultSet} of the rows, or {@code null} on error
+     */
+    public ResultSet query(String query, Object... params) {
+        try (PreparedStatement statement = connection.prepareStatement(query)) {
+            bindParameters(statement, params);
+            try (ResultSet rs = statement.executeQuery()) {
+                CachedRowSet cachedRowSet = RowSetProvider.newFactory().createCachedRowSet();
+                cachedRowSet.populate(rs);
+                return cachedRowSet;
+            }
         } catch (SQLException e) {
             log.error("Error executing query: {}", e.getMessage());
         }
         return null;
     }
 
-    public boolean update(String query) {
-        try {
-            int rowCount = connection.createStatement().executeUpdate(query);
-            return rowCount > 0;
+    /**
+     * Execute a parameterized INSERT/UPDATE/DELETE.
+     *
+     * <p>Run through a {@link PreparedStatement} (parameters bound, not concatenated)
+     * inside a try-with-resources so the statement is always closed.
+     *
+     * @param query  SQL with {@code ?} placeholders for each parameter
+     * @param params values to bind to the placeholders, in order
+     * @return {@code true} if at least one row was affected, {@code false} otherwise (including on error)
+     */
+    public boolean update(String query, Object... params) {
+        try (PreparedStatement statement = connection.prepareStatement(query)) {
+            bindParameters(statement, params);
+            return statement.executeUpdate() > 0;
         } catch (SQLException e) {
             log.error("Error executing update: {}", e.getMessage());
         }
         return false;
+    }
+
+    private void bindParameters(PreparedStatement statement, Object... params) throws SQLException {
+        for (int i = 0; i < params.length; i++) {
+            statement.setObject(i + 1, params[i]);
+        }
     }
 
     public void close() {
--- a/src/main/java/preponderous/viron/factories/EntityFactory.java
+++ b/src/main/java/preponderous/viron/factories/EntityFactory.java
@@ -31,8 +31,8 @@ public class EntityFactory {
             throw new EntityCreationException("Failed to get next entity id");
         }
         String creationDate = new java.util.Date().toString();
-        String query = "INSERT INTO viron.entity (entity_id, name, creation_date) VALUES (" + id + ", '" + name + "', '" + creationDate + "')";
-        boolean success = dbInteractions.update(query);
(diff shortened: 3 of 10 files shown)


dmccoystephenson (commented):
Self-review rubric (adversarial; anchored on green CI run 27468231673 + the diff):

- **Scope:** PASS — 10 files, all necessary: `DbInteractions` rework, the 4 name-bearing call sites (2 factories + 2 repos), their updated test mocks, the new H2 test, and the `pom.xml` test dep. No unrelated churn.
- **Resource fix (#140):** PASS — both methods use try-with-resources on the `PreparedStatement`; `query()` copies rows into a `CachedRowSet` before the live `ResultSet`/statement close. `DbInteractionsTest.query_returnsDisconnectedResultSet_readableAfterReturn` reads the result fully after `query()` returns, which is only possible because it is disconnected.
- **Injection fix (#139):** PASS — **demonstrated, not asserted.** `parameterizedValueWithSqlPayload_isStoredLiterally_andDoesNotInject` inserts `O'Brien'); DROP TABLE person; --` as a bound parameter; the table survives and the value round-trips verbatim. The data layer no longer concatenates user-controlled `name` values.
- **Backward compatibility:** PASS — both signatures are varargs, so the ~40 existing parameterless call sites compile and run unchanged; full suite is 226/226 (was 220). Repositories consume the `CachedRowSet` via the same `ResultSet` API (forward `next()` + `getInt`/`getString` by label/index), all supported.
- **Tests-fix mocks:** PASS — the 12 repository-test stubs whose SQL changed (`?` placeholders + param arg) were updated; `EntityFactoryTest` was updated for the new varargs INSERT.
- **CI:** PASS — `build` green on PR head; CI compiles and runs the H2 test.
- **DTO boundary / Spec / Java-Python parallelism:** N/A — internal data layer; no endpoint, DTO, or Python-client surface changed (the Python client is an HTTP client with no `DbInteractions` mirror).
- **Override correctness:** PASS — only method bodies changed; no `@Override` added or removed.

**Honest caveats for the reviewer:**
1. `query()` now returns a `CachedRowSet` rather than a live `ResultSet` — a deliberate behavioral change in the core data layer (the point of the leak fix). Repositories iterate forward-only, so this is compatible, but it warrants a look.
2. `DbInteractionsTest` covers `getInt`/`getString` by label and the factory's `getInt(1)` by index is standard `CachedRowSet` but not separately asserted here.
3. Integer ids remain string-concatenated (not an injection vector); only user-controlled strings were parameterized to bound the diff.

**This PR is intentionally not auto-merged** — it touches `pom.xml` (do-not-auto-merge). Flagged for manual review.

Issue #149, opened 2026-06-14
Endpoints are unauthenticated over plain HTTP; RFC 0001 specifies JWT auth over HTTPS


dmccoystephenson:
## Summary
RFC 0001 (Preponderous-Software/rfcs, `rfcs/rfc_0001_modular_architecture.md`) states "REST endpoints return JSON and are protected using **JWT authentication over HTTPS**" for all services. Viron currently has **no authentication and no TLS**.

- No Spring Security / JWT artifacts exist: a search for `SecurityConfig`/`jwt`/`bearer`/`spring-boot-starter-security` across `src` and `pom.xml` returns nothing. Every controller endpoint is publicly reachable.
- No TLS: `src/main/resources/application.properties` has `server.port=9999` and **no** `server.ssl.*`; the service runs over plain HTTP.

## Why it matters
Viron stores and mutates simulation/world state. As deployed it is an open read/write API — any caller can enumerate and modify grids, entities, and placements. The RFC's architecture assumes JWT-secured services (UserAuth already issues compatible JWTs).

## Suggested direction
Decide the intended boundary, then either: (a) add JWT bearer-token validation (validating tokens issued by UserAuth) and require auth on mutating endpoints; and/or (b) if Viron is intended to be internal-only behind a gateway, document that explicitly and enforce network isolation. TLS should be terminated at the gateway or configured on the service.

_Filed by Claude on behalf of Daniel Stephenson during an RFC↔implementation alignment audit; claims verified against the Viron source._

PR #160, opened 2026-07-19
fix: attach bearer token to in-repo client SDK requests

### LINKED ISSUE #154: In-repo API clients send no auth token; will 401 against the now-authenticated API (post-#149)
## Summary
Since #149/#150, `SecurityConfig` requires authentication on all endpoints (`.anyRequest().authenticated()`, only `/actuator/health*` and swagger paths are `permitAll`). However, neither in-repo client SDK sends an `Authorization` header, so every non-health call will receive `401 Unauthorized`:
- Java: `src/main/java/preponderous/viron/services/*Service.java` (uses `RestTemplate` with no bearer token).
- Python: `src/main/python/preponderous/viron/services/*Service.py` (uses `requests` with no auth).

## Evidence
- `SecurityConfig.java`: `.anyRequest().authenticated()`.
- `grep -rn "Authorization\|Bearer\|token"` over both `services/` trees returns nothing.

## Proposed work
Give both clients a way to attach a bearer token (e.g. a configurable token / token-provider injected into `RestTemplate` interceptors and the Python `requests` session), and add tests asserting the header is sent.

## Notes
Found during #135 while rebasing onto the #149 auth merge. Pre-existing across all client methods, not introduced by #135.

---

dmccoystephenson:
## Summary
- Since #149/#150 `SecurityConfig` requires authentication on all endpoints except `/actuator/health*` and Swagger paths, but neither in-repo client SDK sent an `Authorization` header, so every non-health call would 401.
- Adds `service.authToken` (Java, `ServiceConfig`) and `auth_token` (Python) as an optional bearer token. A new `AuthTokenInterceptor` attaches `Authorization: Bearer <token>` to every outgoing `RestTemplate` request when a token is configured; the Python SDKs attach the same header via a `get_auth_headers()` helper on each service.
- `EntityService` (Java) already used the shared `RestConfig`-provided `RestTemplate` bean, so it picks up the interceptor automatically; `EnvironmentService`, `GridService`, and `LocationService` each build their own `RestTemplate` via `RestTemplateBuilder` and now register the interceptor explicitly.
- `service.authToken` defaults to empty (`${VIRON_CLIENT_AUTH_TOKEN:}`) — no header is sent unless configured, matching the Python default of no token.

## Test plan
- [x] `./mvnw test -B` — could not run locally (`./mvnw`/`java` unavailable to this session); relying on CI (`.github/workflows/ci.yml`) as the authoritative signal, see below.
- [x] `python3 -m pytest` — 88 passed locally.

Closes #154

---

### DIFF
--- a/src/main/java/preponderous/viron/config/AuthTokenInterceptor.java
+++ b/src/main/java/preponderous/viron/config/AuthTokenInterceptor.java
@@ -1,0 +1,32 @@
+package preponderous.viron.config;
+
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpRequest;
+import org.springframework.http.client.ClientHttpRequestExecution;
+import org.springframework.http.client.ClientHttpRequestInterceptor;
+import org.springframework.http.client.ClientHttpResponse;
+import org.springframework.util.StringUtils;
+
+import java.io.IOException;
+
+/**
+ * Attaches the configured bearer token (service.authToken) to outgoing requests
+ * made by the in-repo client SDKs. If no token is configured, requests are sent
+ * unmodified.
+ */
+public class AuthTokenInterceptor implements ClientHttpRequestInterceptor {
+    private final ServiceConfig serviceConfig;
+
+    public AuthTokenInterceptor(ServiceConfig serviceConfig) {
+        this.serviceConfig = serviceConfig;
+    }
+
+    @Override
+    public ClientHttpResponse intercept(HttpRequest request, byte[] body, ClientHttpRequestExecution execution) throws IOException {
+        String authToken = serviceConfig.getAuthToken();
+        if (StringUtils.hasText(authToken)) {
+            request.getHeaders().set(HttpHeaders.AUTHORIZATION, "Bearer " + authToken);
+        }
+        return execution.execute(request, body);
+    }
+}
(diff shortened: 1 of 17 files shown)


dmccoystephenson:
Self-review: Straightforward, low-risk change. Verified:

1. `AuthTokenInterceptor` only sets the Authorization header when a token is configured (`StringUtils.hasText`), so behavior is unchanged when `service.authToken` is unset — matches the empty default in `application.properties`.
2. `EntityService` picks up the interceptor for free via the shared `RestConfig` `RestTemplate` bean; the other three Java services each register it explicitly since they build their own `RestTemplate` from `RestTemplateBuilder` — consistent with the existing (pre-change) split in how these four services are wired.
3. Python SDK mirrors this with a `get_auth_headers()` helper per service, kept as a plain dict rather than a `requests.Session` to avoid rewriting the existing test-patching pattern (`@patch('requests.get')`, etc.) across all four service/test file pairs.
4. Added dedicated tests: `AuthTokenInterceptorTest` (Java, both branches) and one bearer-token-sends test per Python service, plus updated all pre-existing assertions to include the now-always-present `headers` kwarg.
5. No OpenAPI spec change needed — this only affects the in-repo client SDKs' outbound requests, not the server's public API surface.

Could not run `./mvnw test -B` locally (`mvnw`/`java` blocked in this sandboxed session); `python3 -m pytest` passed 88/88 locally. Relying on CI as the authoritative signal for the Java side.

---

PR #162, opened 2026-07-19
Add location adjacency check, neighbors query, and unoccupied-in-grid query

### LINKED ISSUE #155: feat: entity movement has no adjacency check and no neighbor/reachability query
## Summary

The move endpoint added for the RFC 0001 movement responsibility validates that the entity is placed, that the target location exists, that it is in the same grid, and that it is unoccupied — but it never checks that the target is actually adjacent to the entity's current location. Any location in the grid can currently be reached in a single "move" call, which is a teleport, not a validated movement.

- `src/main/java/preponderous/viron/controllers/LocationController.java:115-135` (`moveEntityToLocation`) — reads the current and target `Location` (which carry `x`/`y`, see `src/main/java/preponderous/viron/models/Location.java:10-12`) but the only checks performed are `currentGrid.equals(targetGrid)` and "target not occupied". There is no comparison of `current.getX()/getY()` against the target's coordinates.
- `src/test/java/preponderous/viron/controllers/LocationControllerTest.java:391-451` — the move test suite (`moveEntityToLocation_Success`, `_DifferentGridIsBadRequest`, `_OccupiedTargetIsConflict`, etc.) covers grid-mismatch and occupancy, but has no test for a non-adjacent target being rejected, confirming the gap is not just untested but genuinely unvalidated in the controller logic.
- The suggested direction on the now-closed collision/occupancy issue (Preponderous-Software/viron#147) explicitly called out `"neighbors of location"` as a query to add alongside `entities at location` / `is location occupied` — only the latter two were implemented (`getEntityIdsAtLocation`, `isLocationOccupied` in `LocationController.java:94-108`); no neighbor/adjacency query exists (`grep -rn "neighbor" src/main` returns nothing).

## Why it matters

RFC 0001 (Preponderous-Software/rfcs, `rfcs/rfc_0001_modular_architecture.md`) frames Viron's movement responsibility around a game like Ophidian, where "end game if collision with wall/self" depends on step-by-step, legal movement. As implemented, a caller can move an entity from any location to any other location in the same grid in one call, which defeats step-based movement rules a consuming game would rely on, and there is no way to ask Viron "what locations can this entity legally move to next" without fetching every location in the grid and probing occupancy one-by-one.

## Suggested fix

- Add an adjacency check to `moveEntityToLocation` (e.g. Chebyshev or Manhattan distance of 1 between `current.getX()/getY()` and the target's coordinates, matching the grid's coordinate scheme already used by `Location`), rejecting non-adjacent moves the same way the existing different-grid case is rejected (`InvalidRequestException`).
- Add a `GET /api/v1/locations/{locationId}/neighbors` endpoint (mirroring the existing `GET /api/v1/locations/{locationId}/entities` and `.../occupied` pattern) that returns the adjacent locations within the same grid, so a caller can discover legal move targets without brute-forcing the grid.

_Filed by Claude during an automated feature-opportunity triage pass; claims above were verified against source._

### LINKED ISSUE #159: feat: no way to query unoccupied locations in a grid (mirrors the existing entity 'unassigned' pattern)
## Summary

`EntityController` already establishes the pattern of a dedicated query for "things not currently placed": `GET /api/v1/entities/unassigned` returns entities that aren't in any location. `LocationController` has the mirror-image query for entities within a single location (`GET /{locationId}/entities`) and a single-location occupancy check (`GET /{locationId}/occupied`), but there is no equivalent bulk query for "which locations in this grid are free" — a caller must fetch every location in a grid and then call `/occupied` once per location to find an open spot.

- `src/main/java/preponderous/viron/controllers/EntityController.java:63-67` — `getEntitiesNotInAnyLocation()` → `GET /unassigned`, backed by `EntityRepository.findEntitiesNotInAnyLocation()` (`src/main/java/preponderous/viron/repositories/EntityRepositoryImpl.java:132-148`), a single `NOT IN` subquery.
- `src/main/java/preponderous/viron/controllers/LocationController.java:94-108` — `getEntityIdsAtLocation`/`isLocationOccupied` only answer the question for one location at a time; there is no `LocationRepository` method that returns locations in a grid filtered by occupancy (`grep -n "unoccupied\|available\|vacant\|free" src/main/java/preponderous/viron/repositories/LocationRepository*.java` returns nothing).
- The newly added move endpoint (`LocationController.java:135-152`) specifically needs an unoccupied target, making "which nearby/grid locations are currently free" a query a caller of that endpoint would naturally need next.

## Why it matters

Without a bulk query, finding a free location in an N-location grid costs N+1 round trips (one `GET /locations/grid/{gridId}` plus one `GET /{locationId}/occupied` per location) instead of one. This is the same class of gap the entity `/unassigned` endpoint already exists to avoid on the entity side, so `DebugController`'s own placement logic (`src/main/java/preponderous/viron/controllers/DebugController.java`, `createSampleData`/`createWorldAndPlaceEntity`) and any future movement-planning caller are left re-deriving it via brute force.

## Suggested fix

Add `GET /api/v1/locations/grid/{gridId}/unoccupied` to `LocationController`, backed by a new `LocationRepository` method (e.g. a `NOT IN` subquery against `entity_location` scoped to the grid's locations, following the same shape as `EntityRepositoryImpl.findEntitiesNotInAnyLocation`), returning the `LocationDto` list of free locations in that grid.

_Filed by Claude during an automated feature-opportunity triage pass; claims above were verified against source._

dmccoystephenson:
## Summary
- `moveEntityToLocation` now rejects a target location that isn't adjacent (Chebyshev distance 1, including diagonals) to the entity's current location, turning the previously-any-location "move" into real step-based movement validation.
- Add `GET /api/v1/locations/{locationId}/neighbors` to discover adjacent locations in the same grid without brute-forcing every location + an occupancy check per candidate.
- Add `GET /api/v1/locations/grid/{gridId}/unoccupied` to find free locations in a grid in a single call, mirroring the existing `EntityController` "unassigned" pattern.
- Update `docs/openapi/viron-api.json` and the Python SDK (`LocationService`) to match the new/changed endpoints.

## Test plan
- [x] `mvn test -B` — full suite green (ran via `mvn` directly; `./mvnw` itself required interactive approval in this sandboxed environment, `mvn` on PATH is equivalent for this repo)
- [x] `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -m pytest` — 98 passed (the ambient `pytest-cov`/`pytest` version mismatch in this environment crashes plugin autoload; disabling autoload works around it, unrelated to this change)
- [x] CI (`.github/workflows/ci.yml`) is the authoritative signal — will confirm green before requesting merge

Closes #155
Closes #159

---

### DIFF
--- a/docs/openapi/viron-api.json
+++ b/docs/openapi/viron-api.json
@@ -175,6 +175,45 @@
         }
       }
     },
+    "/api/v1/locations/grid/{gridId}/unoccupied": {
+      "get": {
+        "summary": "Get unoccupied locations in a grid",
+        "parameters": [{ "$ref": "#/components/parameters/GridIdParam" }],
+        "responses": {
+          "200": {
+            "description": "List of unoccupied locations in the grid",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "array",
+                  "items": { "$ref": "#/components/schemas/LocationDTO" }
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "/api/v1/locations/{locationId}/neighbors": {
+      "get": {
+        "summary": "Get the locations adjacent to a location within the same grid",
+        "parameters": [{ "$ref": "#/components/parameters/LocationIdParam" }],
+        "responses": {
+          "200": {
+            "description": "List of adjacent locations",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "array",
+                  "items": { "$ref": "#/components/schemas/LocationDTO" }
+                }
+              }
+            }
+          },
+          "404": { "description": "Location not found, or not in any grid" }
+        }
+      }
+    },
     "/api/v1/locations/{locationId}/entity/{entityId}": {
       "put": {
         "summary": "Add entity to location",
@@ -195,14 +234,14 @@
     },
     "/api/v1/locations/{locationId}/entity/{entityId}/move": {
       "put": {
-        "summary": "Move an entity to an unoccupied location in the same grid",
+        "summary": "Move an entity to an adjacent, unoccupied location in the same grid",
         "parameters": [
           { "$ref": "#/components/parameters/LocationIdParam" },
           { "$ref": "#/components/parameters/EntityIdParam" }
         ],
         "responses": {
           "204": { "description": "Entity moved" },
-          "400": { "description": "Target location is not in the same grid as the entity" },
+          "400": { "description": "Target location is not in the same grid as the entity, or is not adjacent to the entity's current location" },
           "404": { "description": "Entity is not placed at any location, or target location not found" },
           "409": { "description": "Target location is already occupied" }
         }
--- a/src/main/java/preponderous/viron/controllers/LocationController.java
+++ b/src/main/java/preponderous/viron/controllers/LocationController.java
@@ -17,6 +17,7 @@ import preponderous.viron.repositories.LocationRepository;
 import jakarta.validation.constraints.Min;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 @RestController
 @RequestMapping("/api/v1/locations")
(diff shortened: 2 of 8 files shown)


dmccoystephenson:
Self-review: no blocking issues found — see PR description for what changed. Adjacency check reuses existing InvalidRequestException/NotFoundException (no new exception wiring), no new DTOs, no security-surface change (both new endpoints sit under the already-authenticated /api/v1/locations/** path). OpenAPI spec and Python SDK updated in this same PR. Full mvn test -B and pytest suite pass locally; CI is the authoritative signal.

---

PR #167, opened 2026-07-25
feat: support non-square grids and gate debug endpoints behind viron.debug.enabled

### LINKED ISSUE #156: feat: debug endpoints have no profile/flag gate and are always live
## Summary

`docs/REBUILD_PLAN.md` explicitly says debug tooling should not ship unconditionally to production, and `docs/PLANNING.md`'s M5 milestone lists "Operational debug toggles — Enable/disable debug endpoints" as a planned item. Neither exists: `DebugController` is a plain, always-registered `@RestController` with no gating.

- `docs/REBUILD_PLAN.md` (Notes section) — "Keep debug tools behind `spring.profiles.active=debug` so they don't ship to production."
- `docs/PLANNING.md`, milestone M5, item 60 — "Operational debug toggles — Enable/disable debug endpoints."
- `src/main/java/preponderous/viron/controllers/DebugController.java:25-29` — `@RestController @RequestMapping("/api/v1/debug") @Slf4j @Validated` with no `@Profile`, `@ConditionalOnProperty`, or any other conditional annotation.
- `src/main/java/preponderous/viron/config/SecurityConfig.java:44-54` — the security filter chain treats `/api/v1/debug/**` the same as every other endpoint (`anyRequest().authenticated()`); there is no separate carve-out or extra restriction for it.
- `src/main/resources/application.properties` — no `spring.profiles.active` setting and no `debug.enabled`-style property anywhere in the file or in `compose.yml`.

## Why it matters

`POST /api/v1/debug/create-sample-data` and `POST /api/v1/debug/create-world-and-place-entity/{environmentName}` create arbitrary environments, grids, locations, and entities. Any authenticated caller — which, per #149, is now any holder of a valid UserAuth-issued JWT, not necessarily an operator of this deployment — can invoke them in any environment the service is running in, including production, with no way for an operator to disable them short of removing the controller from the build.

## Suggested fix

Gate `DebugController` behind a configuration switch consistent with the rebuild plan's own stated intent, e.g.:
- Add a `viron.debug.enabled` property (default `false`, matching Spring Boot's `@ConditionalOnProperty(prefix = "viron.debug", name = "enabled", havingValue = "true")` on the controller class), or
- Restrict it to a Spring profile (`@Profile("debug")`) activated only in dev/local `compose.yml`/`application-debug.properties`, per the REBUILD_PLAN.md note.

Either approach keeps the existing debug endpoints for local/dev use while preventing them from being reachable by default in a production deployment.

_Filed by Claude during an automated feature-opportunity triage pass; claims above were verified against source._

### LINKED ISSUE #158: feat: environment creation cannot produce non-square grids the data model already supports
## Summary

`Grid`, its DTO, repository, and database table all model rows and columns as independent values, but the only way to create a grid — creating an environment — forces every grid to be square by reusing a single `gridSize` value for both dimensions.

- `src/main/java/preponderous/viron/models/Grid.java:11-12` — `Grid(int gridId, int rows, int columns)`, independent fields.
- `src/main/java/preponderous/viron/dto/GridDto.java` — separate `rows` and `columns` properties, each independently documented via `@Schema`.
- `db-scripts/setup/create_tables.sql:22-26` — `viron.grid` has independent `rows INT NOT NULL` and `columns INT NOT NULL` columns with no constraint tying them together.
- `src/main/java/preponderous/viron/dto/CreateEnvironmentRequest.java:19-25` — the only creation entry point takes a single `gridSize` (`@Min(1) private int gridSize;` — "Size (rows and columns) of each grid").
- `src/main/java/preponderous/viron/factories/EnvironmentFactory.java:55` — `"INSERT INTO viron.grid (grid_id, rows, columns) VALUES (" + nextGridId + ", " + gridSize + ", " + gridSize + ")"` — the same value is inserted for both `rows` and `columns`, and the nested location-creation loop at lines 72-92 iterates `x` and `y` both from `0` to `gridSize`, so every generated location grid is also forced square.

## Why it matters

The README frames Viron as letting a caller "spawn an entire planet" via one `POST /api/v1/environments` call, and the MVP scope explicitly lists grid rows/columns as independently modeled data. A caller who wants a non-square board (e.g. a wide corridor, a rectangular map, or anything that isn't literally N×N) cannot express that through the API today even though nothing downstream — the model, DTO, repository, or schema — actually requires grids to be square. The constraint exists only in the request DTO and factory, not the data it's writing.

## Suggested fix

Add `numRows`/`numColumns` (or `gridWidth`/`gridHeight`) fields to `CreateEnvironmentRequest` alongside (or replacing) `gridSize`, thread them through `EnvironmentFactory.createEnvironment` in place of the duplicated `gridSize` argument (`INSERT INTO viron.grid (...) VALUES (..., numRows, numColumns)`), and adjust the location-generation loop to iterate `x` in `[0, numRows)` and `y` in `[0, numColumns)` independently. For backward compatibility, `gridSize` could remain as an optional convenience field that sets both dimensions when the new fields are omitted.

_Filed by Claude during an automated feature-opportunity triage pass; claims above were verified against source._

dmccoystephenson:
## Summary

Two independent-but-adjacent changes to environment creation and the debug surface.

**Non-square grids (#158)** — `Grid`, `GridDto`, and `viron.grid` have always modelled `rows` and `columns` independently, but the only creation path forced them equal.

- `CreateEnvironmentRequest` gains `numRows`/`numColumns`; `gridSize` becomes an optional square shorthand (`Integer`, so its absence is distinguishable from `0`).
- An `@AssertTrue` guard rejects a half-specified request (only `numRows`, or only `numColumns`) instead of silently falling back to a square grid — so `gridSize` alone, or both dimensions, are the only accepted shapes.
- `EnvironmentFactory.createEnvironment` now takes `(name, numGrids, numRows, numColumns)` and iterates `x` over rows and `y` over columns independently. The grid `INSERT` is parameterized while it was being rewritten, matching the `DbInteractions` hardening from #143.
- A 3-arg convenience constructor keeps `new CreateEnvironmentRequest(name, numGrids, gridSize)` compiling, so the existing square call sites (client SDK, `DebugController`) are untouched.

**Debug endpoint gating (#156)** — `docs/REBUILD_PLAN.md` says debug tools shouldn't ship to production and `docs/PLANNING.md` M5 lists "operational debug toggles"; neither existed.

- `DebugController` is now `@ConditionalOnProperty(prefix = "viron.debug", name = "enabled", havingValue = "true")`. Absent or `false`, the bean isn't registered and `/api/v1/debug/**` isn't mapped at all — so data-seeding is no longer reachable by every holder of a valid UserAuth JWT.
- `viron.debug.enabled=${VIRON_DEBUG_ENABLED:false}` in `application.properties`, plumbed through `compose.yml` and `sample.env`.

**Kept in sync:**
- `docs/openapi/viron-api.json` — `CreateEnvironmentRequest` schema (`gridSize` dropped from `required`, `numRows`/`numColumns` added, semantics described); both debug paths note the flag. The existing Postman example body still sends `name`/`numGrids`/`gridSize`, which remains valid, so `postman/` needs no regeneration.
- Python SDK — `create_environment` takes optional `grid_size`/`num_rows`/`num_columns` and mirrors the server's "fully specified dimensions" rule client-side with a `ValueError`.
- `README.md` and `docs/MVP.md` — new request shape and the default-off debug flag.

## Test plan

- [x] `python3 -m pytest` — 101 passed (adds 3 cases for the new `create_environment` signature).
- [ ] `./mvnw test -B` — **not runnable in this environment** (`java`/`mvn` are permission-denied in the sandbox this loop ran in). Relying on CI's `./mvnw test -B` as the authoritative signal; see the checks on this PR.

New/updated Java tests:
- `CreateEnvironmentRequestTest` (new) — dimension resolution and the validity predicate, including precedence and half-specified rejection.
- `EnvironmentControllerTest` — rectangular create, `numRows`/`numColumns` precedence over `gridSize`, and 400s for no dimensions / only `numRows` / non-positive `numColumns`.
- `DebugControllerDisabledTest` (new) — with no `viron.debug.enabled`, no `debugController` bean and no `/api/v1/debug` handler mapping. Asserts against `RequestMappingHandlerMapping` rather than an HTTP status, because `GlobalExceptionHandler`'s catch-all `@ExceptionHandler(Exception.class)` intercepts unmapped-path exceptions, making the resulting status an unreliable signal.
- `DebugControllerTest` — `@TestPropertySource(properties = "viron.debug.enabled=true")` so the existing coverage still exercises the controller.

## Notes

`viron.debug.enabled` defaults to **off**, so any deployment currently relying on `/api/v1/debug/**` must set `VIRON_DEBUG_ENABLED=true`. That is the intent of #156, but it is a behavioural break worth calling out.

Closes #158
Closes #156

---


dmccoystephenson:
## Self-review

Checked against this repo's conventions: layering (DTO → factory → controller), Bean Validation + `@Schema` on every request-DTO field, no new exception types (so no `GlobalExceptionHandler` gap), parameterized SQL and no leaked JDBC resources in the touched `INSERT`, OpenAPI spec updated in the same PR, and the Python SDK updated to match the changed request contract — the drift this repo most often accumulates.

**No blocking findings.** Four notes.

---

**`src/main/java/preponderous/viron/dto/CreateEnvironmentRequest.java:58,70` — unboxing NPE on an unvalidated instance**

`getResolvedNumRows()`/`getResolvedNumColumns()` unbox `gridSize` and will NPE when every dimension field is null. That is unreachable on the request path — `@Valid` runs `isGridDimensionsSpecified()` before the handler body, and the only other caller is the client SDK, which always supplies one form or the other. The javadoc says as much; flagging it so it reads as a conscious constraint rather than an accident if someone later reuses this DTO outside a `@Valid` parameter.

**`src/main/java/preponderous/viron/dto/CreateEnvironmentRequest.java:79` — the 400 body names a synthetic property**

Because this is a getter-level constraint, `GlobalExceptionHandler.handleValidationException` renders the 400 as `gridDimensionsSpecified: either gridSize or both numRows and numColumns must be provided`. The message is actionable, but the field prefix names a property that has no counterpart in the request body, which reads oddly for an API consumer. Acceptable as-is — the alternative (a class-level custom constraint) adds an annotation + validator pair for a cosmetic gain — but that is the exact 400 body callers will see.

**`src/main/java/preponderous/viron/factories/EnvironmentFactory.java:64` — partial SQL parameterization**

Parameterized this `INSERT` while rewriting it for independent dimensions, per the #139/#140 convention. The remaining concatenated `INSERT`s in the same method (`grid_environment`, `location`, `location_grid`) are untouched: every interpolated value there is an `int` from a sequence getter or a loop counter, so there is no injection surface, and rewriting them would widen this PR past the two issues it closes. Worth a follow-up for consistency, not a defect.

**`src/main/java/preponderous/viron/controllers/DebugController.java:37` — behavioural break for running deployments**

`/api/v1/debug/**` goes from always-on to off-unless-`VIRON_DEBUG_ENABLED=true`. That is precisely what #156 asks for, and `compose.yml` / `sample.env` / `README.md` / `docs/MVP.md` all document the switch, but it is the one change here that could surprise a running environment rather than just a caller.

---

**Deliberately left alone:** `postman/Viron.postman_collection.json` still sends `{name, numGrids, gridSize}` for `POST /api/v1/environments`, which remains valid under the new schema. Regenerating it would derive the example body from `required`, which no longer includes `gridSize`, producing an example that fails validation — so not regenerating is the better of the two options here.

**Verification caveat:** `./mvnw test -B` was not run locally — `java`/`mvn` are permission-denied in the sandbox this dev loop ran in, so CI's `build` job is the authoritative signal for the Java half. The Python half was verified locally (101 passed).

_(Posted as a comment rather than a formal review with anchored inline comments: the review API is not reachable from this session.)_

---

Issue #130, opened 2026-06-13
Wire OpenAPI-first codegen / spec-drift guard (rebuild-plan hygiene gap)


dmccoystephenson:
## Summary
Two hygiene mechanisms prescribed by `docs/REBUILD_PLAN.md` were not adopted, which leaves spec/code alignment maintained by hand and allows drift (see the write-endpoint divergence tracked separately).

## Gaps vs. the rebuild plan
1. **OpenAPI-first codegen is not wired in.** Steps 4 & 6 of the plan call for an `openapi-generator` build step that produces DTOs/interfaces from `docs/openapi/viron-api.json`. `pom.xml` contains no generator; DTOs and controllers are hand-written. Without it, nothing enforces that the implementation matches the contract.
2. **No spec-drift guard in CI.** There is no check that fails when controllers and `viron-api.json` disagree.

## Proposed work
- Evaluate wiring `openapi-generator-maven-plugin` to generate request/response DTOs (or interfaces) from the spec, or alternatively generate the spec *from* the code via springdoc and diff it in CI.
- Add a CI step that fails on spec/implementation drift.

## Notes
This is a developer-experience / quality-gate improvement rather than user-facing functionality, and may be deferred; it is filed so the deviation from the rebuild plan is tracked.

---
*Filed by Claude on behalf of Daniel Stephenson.*

dmccoystephenson:
Deferred from the 2026-06-13 dev-loop cycle (which shipped #133, the EnvironmentRepositoryImpl tests). Skip reason: this is a larger architectural devexp change touching the `pom.xml` build pipeline (wiring openapi-generator / a CI spec-drift gate); the issue itself notes it 'may be deferred.' It is beyond a single-cycle scope ceiling and is not a quick win — leaving it for a dedicated cycle.

dmccoystephenson:
Still deferred this cycle (2026-07-28, PR #177 addressed #148 instead) for the same reason as before: a larger architectural devexp change touching the `pom.xml` build pipeline, explicitly noted as possibly-deferrable in the issue body, and not a quick win relative to a single-cycle scope ceiling.

---

_drafted by Claude on behalf of Daniel Stephenson_

dmccoystephenson:
Partial progress in PR #179: added a CI spec-drift guard (a full-context Spring Boot test that compares springdoc's live /v3/api-docs route list against docs/openapi/viron-api.json and fails the build on mismatch). This closes the "no spec-drift guard in CI" gap.

[...]

PR #173, opened 2026-07-26
docs: align OpenAPI error responses with the implementation, and fix SDK delete_environment

### LINKED ISSUE #169: docs: OpenAPI spec documents error statuses with no body schema, but the API returns ErrorResponse
## Summary

`docs/openapi/viron-api.json` documents error responses with a `description` and no body schema, e.g.:

```json
"404": { "description": "Location not found" }
```

But the API does return a body. `exceptions/GlobalExceptionHandler.java` is a `@RestControllerAdvice` whose handlers all return `dto/ErrorResponse`, which carries two fields:

```java
public class ErrorResponse {
    private int status;
    private String message;
}
```

All ten `@ExceptionHandler` methods on the advice return `ResponseEntity<ErrorResponse>` — including the `MethodArgumentNotValidException` and catch-all `Exception` handlers — so the shape is uniform across every failure mode.

`ErrorResponse` is the only DTO in `src/main/java/preponderous/viron/dto/` with no counterpart in the spec's `components/schemas` (the other eight — `EnvironmentDto`, `GridDto`, `LocationDto`, `EntityDto`, `CreateEnvironmentRequest`, `UpdateEnvironmentNameRequest`, `CreateEntityRequest`, `UpdateEntityNameRequest` — are all present). There are currently 10 error responses across the spec (8× `404`, 1× `400`, 1× `409`), none of which declares content.

## Why it matters

A client generated from this spec — the stated goal of #130 — gets no type for the error payload and has to hand-roll one or fall back to parsing `object`. `postman/` is generated from the spec too, so the collection's saved responses have nothing to show for the failure cases. It also means the spec understates the API: a consumer reading it would reasonably conclude a 404 has an empty body.

## Suggested fix

1. Add an `ErrorResponse` schema to `components/schemas`, matching the two fields on the Java DTO.
2. Give each existing `4xx`/`409` response a `content.application/json.schema.$ref` pointing at it.

Both steps are mechanical, but step 2 touches every error response in the file, which is why it was left out of #168 (the doc-alignment PR that added the four missing paths) rather than folded in.

Note there is no dedicated `GlobalExceptionHandler` test class — the serialized shape is asserted indirectly through the controller tests' `jsonPath("$.status")` / `jsonPath("$.message")` expectations. Adding a direct test for the advice would be a reasonable companion change, but is not required for the spec fix.

### LINKED ISSUE #170: docs: OpenAPI spec documents the wrong success status for five endpoints (three DELETEs and both debug POSTs)
## Summary

Five endpoints in `docs/openapi/viron-api.json` document a success status that the controller does not actually return.

| Endpoint | Spec says | Controller returns |
|---|---|---|
| `DELETE /api/v1/environments/{id}` | `200` | `204` — `EnvironmentController.java:73-75` `@ResponseStatus(HttpStatus.NO_CONTENT)` |
| `DELETE /api/v1/locations/{locationId}/entity/{entityId}` | `200` | `204` — `LocationController.java:79-81` `@ResponseStatus(HttpStatus.NO_CONTENT)` |
| `DELETE /api/v1/locations/entity/{entityId}` | `200` | `204` — `LocationController.java:90-92` `@ResponseStatus(HttpStatus.NO_CONTENT)` |
| `POST /api/v1/debug/create-sample-data` | `200` | `201` — `DebugController.java:65-67` `@ResponseStatus(HttpStatus.CREATED)` |
| `POST /api/v1/debug/create-world-and-place-entity/{environmentName}` | `200` | `201` — `DebugController.java:106-108` `@ResponseStatus(HttpStatus.CREATED)` |

For contrast, the endpoints that *are* correct: `DELETE /api/v1/entities/{id}` and `PUT /api/v1/locations/{locationId}/entity/{entityId}/move` are both documented as `204` and both carry `@ResponseStatus(HttpStatus.NO_CONTENT)`; `POST /api/v1/entities` and `POST /api/v1/environments` are documented as `201` and both carry `@ResponseStatus(HttpStatus.CREATED)`. The convention is already established — these five are simply out of step with it.

`PUT /api/v1/locations/{locationId}/entity/{entityId}` (add entity, no `@ResponseStatus`) and both `PATCH .../name` endpoints return `void` with no `@ResponseStatus`, which Spring serves as `200` — those match the spec and need no change.

## Why it matters

A client generated from this spec (the goal of #130) will treat a `204` as an unexpected status, and the `postman/` collection — generated from the spec — asserts the wrong code. The two debug endpoints are the ones most likely to be hit first by someone kicking the tires on the API.

## Suggested fix

Rename the response key (`200` → `204` / `201`) for the five entries above. The three `204` cases have no body and no `content` block today, so the rename is the whole change; the two debug `201` cases keep their existing `content` block with the `EnvironmentDTO` / `EntityDTO` `$ref`.

_Filed by Claude during an automated triage pass; every claim above was verified against source._

---

### LINKED ISSUE #171: docs: 13 documented-in-code failure modes have no response entry in the OpenAPI spec (11 missing 404s, 2 missing 400s)
## Summary

`docs/openapi/viron-api.json` documents `404` on only 8 of the endpoints that can return one, and documents no `400` for the two debug endpoints that can return one. Thirteen documented failure modes are missing in total.

### Missing `404` (controller throws `NotFoundException`, which `GlobalExceptionHandler` maps to `404`)

| Endpoint | Throw site |
|---|---|
| `GET /api/v1/environments/{id}` | `EnvironmentController.java:46` |
| `DELETE /api/v1/environments/{id}` | `EnvironmentController.java:77` |
| `PATCH /api/v1/environments/{id}/name` | `EnvironmentController.java:147` |
| `GET /api/v1/grids/{id}` | `GridController.java:34` |
| `GET /api/v1/locations/{id}` | `LocationController.java:40` |
| `PUT /api/v1/locations/{locationId}/entity/{entityId}` | `LocationController.java:72` |
| `DELETE /api/v1/locations/{locationId}/entity/{entityId}` | `LocationController.java:83` |
| `DELETE /api/v1/locations/entity/{entityId}` | `LocationController.java:94` |
| `GET /api/v1/entities/{id}` | `EntityController.java:41` |
| `DELETE /api/v1/entities/{id}` | `EntityController.java:80` |
| `PATCH /api/v1/entities/{id}/name` | `EntityController.java:91` |

Note `GET /api/v1/locations/entity/{entityId}` already documents its `404`, but the `DELETE` on the same path does not — the two operations sit side by side in the spec, so this one is easy to miss.

### Missing `400` (controller throws `InvalidRequestException`, mapped to `400`)

| Endpoint | Throw sites |
|---|---|
| `POST /api/v1/debug/create-sample-data` | `DebugController.java:72`, `:92` |
| `POST /api/v1/debug/create-world-and-place-entity/{environmentName}` | `DebugController.java:118`, `:140` |

## Why it matters

Same consumer impact as #169: a generated client (#130) and the generated `postman/` collection only know about the failure modes the spec declares. Here the spec doesn't merely understate the *body* of an error — it claims a whole class of endpoints can't fail at all, when e.g. `GET /api/v1/entities/{id}` for a missing id is one of the most ordinary error cases in the API.

## Scope note

Deliberately **not** included: the `400` that every `@Min(1)` path variable and every `@Valid` request body can produce via `HandlerMethodValidationException` / `MethodArgumentNotValidException`. That applies to nearly every endpoint in the API and is better handled as one uniform sweep (or by generating the spec from the code, per #130) than as a hand-edit here. The 13 above are the failure modes each controller raises explicitly in its own body.

## Suggested fix

Add the missing response entries. Best done together with #169, which adds the `ErrorResponse` schema these entries should `$ref`, and #170, which corrects the success codes on four of the same operations.

_Filed by Claude during an automated triage pass; every claim above was verified against source._

---

### LINKED ISSUE #172: bug: Python SDK delete_environment returns False on success (checks for 200, API returns 204)
## Summary

`EnvironmentService.delete_environment` in the Python SDK returns `False` after a **successful** delete.

`src/main/python/preponderous/viron/services/environmentService.py:66-69`:

```python
def delete_environment(self, environment_id: int) -> bool:
    response = requests.delete(f"{self.get_base_url()}/{environment_id}", headers=self.get_auth_headers())
    response.raise_for_status()
    return response.status_code == 200
```

But `EnvironmentController.deleteEnvironment` (`EnvironmentController.java:73-75`) is annotated `@ResponseStatus(HttpStatus.NO_CONTENT)` and returns **204**. `raise_for_status()` passes (204 is not an error), then `status_code == 200` evaluates `False`. So the happy path reports failure, and the only way the method returns `True` is a status the API never sends.

## Why the test suite doesn't catch it

`src/test/python/preponderous/viron/services/test_environmentService.py:153-162` mocks the response as `status_code = 200`, i.e. the test asserts against the same incorrect assumption the implementation makes rather than against the real contract. The SDK also has no CI coverage — `.github/workflows/ci.yml` only runs `./mvnw compile` and `./mvnw test` — so nothing else flags it either.

For contrast, `EntityService.delete_entity` (`entityService.py:65-72`) does this correctly: it calls `raise_for_status()` and then returns `True` unconditionally, so it is agnostic to which 2xx the API picks.

`EnvironmentService.update_environment_name` also ends in `return response.status_code == 200`, but `PATCH /api/v1/environments/{id}/name` genuinely returns 200 (`void` handler with no `@ResponseStatus`), so that one is correct as written.

## Suggested fix

Make `delete_environment` mirror `delete_entity` — return `True` once `raise_for_status()` has passed — and fix the mocked status in `test_delete_environment` to 204 so the test pins the real contract.

Related: #170 (the OpenAPI spec documents this same endpoint as 200 for the same reason).

_Filed by Claude during an automated triage pass; every claim above was verified against source._

dmccoystephenson:
## Summary

Brings `docs/openapi/viron-api.json` in line with what the controllers and `GlobalExceptionHandler` actually do, and fixes one Python SDK bug that shared a root cause with the spec drift.

**Error bodies (#169)** — adds an `ErrorResponse` schema (`status`, `message`, matching `dto/ErrorResponse.java`) and references it from all 23 `4xx` responses. Previously every error response declared a `description` and no body, so the spec claimed a 404 came back empty.

**Wrong success statuses (#170)** — five operations documented a status their controller does not return:

| Endpoint | Was | Now |
|---|---|---|
| `DELETE /api/v1/environments/{id}` | 200 | 204 |
| `DELETE /api/v1/locations/{locationId}/entity/{entityId}` | 200 | 204 |
| `DELETE /api/v1/locations/entity/{entityId}` | 200 | 204 |
| `POST /api/v1/debug/create-sample-data` | 200 | 201 |
| `POST /api/v1/debug/create-world-and-place-entity/{environmentName}` | 200 | 201 |

**Undocumented failure modes (#171)** — adds the 11 missing `404`s (every place a controller throws `NotFoundException`) and the 2 missing `400`s on the debug endpoints (`InvalidRequestException`). Deliberately excluded, per that issue's scope note: the `400` that `@Min(1)` / `@Valid` can produce on nearly every endpoint — better done as one uniform sweep or via #130.

**Python SDK (#172)** — `EnvironmentService.delete_environment` returned `response.status_code == 200` for an endpoint that returns 204, so a successful delete reported `False`. It now returns `True` once `raise_for_status()` has passed, matching `EntityService.delete_entity`. Its test mocked `status_code = 200`, which is why the suite never caught it; the mock is now 204.

`docs/MVP.md`'s DTO checklist gains an `ErrorResponse` line, since the schema is now part of the documented contract.

No Java source changed.

## Test plan

- [x] `python3 -m pytest` — 101 passed
- [x] `docs/openapi/viron-api.json` parses as JSON; every `$ref` in the document resolves; every `4xx`/`5xx` response now carries `content.application/json.schema.$ref → ErrorResponse` (checked with a throwaway script, not committed)
- [x] Documented statuses diffed operation-by-operation against `@ResponseStatus` annotations and `throw new *Exception` sites in all five controllers
- [ ] `./mvnw test -B` — not runnable in this environment (no local JDK/Maven); relying on the CI workflow's `./mvnw compile -B` + `./mvnw test -B` on this branch. No Java files are touched, so this is a no-op regression check.

Closes #169
Closes #170
Closes #171
Closes #172

---

### DIFF
--- a/docs/MVP.md
+++ b/docs/MVP.md
@@ -89,6 +89,7 @@ Deliver a working, tested API that supports:
 - [x] **EntityDTO** – Public representation of an entity.
 - [x] **CreateEntityRequest** – Request body for creating entities (name).
 - [x] **UpdateEntityNameRequest** – Request body for updating entity names.
+- [x] **ErrorResponse** – Body returned for every 4xx and 5xx response (`status`, `message`).
 
 ---
 
--- a/docs/openapi/viron-api.json
+++ b/docs/openapi/viron-api.json
@@ -51,13 +51,27 @@
             "content": {
               "application/json": { "schema": { "$ref": "#/components/schemas/EnvironmentDTO" } }
             }
+          },
+          "404": {
+            "description": "Environment not found with that id",
+            "content": {
+              "application/json": { "schema": { "$ref": "#/components/schemas/ErrorResponse" } }
+            }
           }
         }
       },
       "delete": {
         "summary": "Delete environment",
         "parameters": [{ "$ref": "#/components/parameters/EnvironmentIdParam" }],
-        "responses": { "200": { "description": "Environment deleted" } }
+        "responses": {
+          "204": { "description": "Environment deleted" },
+          "404": {
+            "description": "Environment not found with that id",
+            "content": {
+              "application/json": { "schema": { "$ref": "#/components/schemas/ErrorResponse" } }
+            }
+          }
+        }
       }
     },
     "/api/v1/environments/name/{name}": {
@@ -78,7 +92,12 @@
               "application/json": { "schema": { "$ref": "#/components/schemas/EnvironmentDTO" } }
             }
           },
-          "404": { "description": "Environment not found with that name" }
+          "404": {
+            "description": "Environment not found with that name",
+            "content": {
+              "application/json": { "schema": { "$ref": "#/components/schemas/ErrorResponse" } }
+            }
+          }
         }
       }
     },
@@ -93,7 +112,12 @@
               "application/json": { "schema": { "$ref": "#/components/schemas/EnvironmentDTO" } }
             }
           },
-          "404": { "description": "Environment not found for entity" }
+          "404": {
+            "description": "Environment not found for entity",
+            "content": {
+              "application/json": { "schema": { "$ref": "#/components/schemas/ErrorResponse" } }
+            }
+          }
         }
       }
     },
@@ -107,7 +131,15 @@
             "application/json": { "schema": { "$ref": "#/components/schemas/UpdateEnvironmentNameRequest" } }
           }
         },
-        "responses": { "200": { "description": "Name updated" } }
+        "responses": {
+          "200": { "description": "Name updated" },
+          "404": {
+            "description": "Environment not found with that id",
+            "content": {
+              "application/json": { "schema": { "$ref": "#/components/schemas/ErrorResponse" } }
+            }
+          }
+        }
       }
     },
     "/api/v1/grids": {
@@ -137,6 +169,12 @@
             "description": "Grid details",
             "content": {
               "application/json": { "schema": { "$ref": "#/components/schemas/GridDTO" } }
+            }
+          },
+          "404": {
+            "description": "Grid not found with that id",
+            "content": {
+              "application/json": { "schema": { "$ref": "#/components/schemas/ErrorResponse" } }
             }
           }
         }
@@ -179,7 +217,12 @@
               "application/json": { "schema": { "$ref": "#/components/schemas/GridDTO" } }
             }
           },
-          "404": { "description": "Grid not found for entity" }
+          "404": {
+            "description": "Grid not found for entity",
+            "content": {
+              "application/json": { "schema": { "$ref": "#/components/schemas/ErrorResponse" } }
+            }
+          }
         }
       }
     },
@@ -210,6 +253,12 @@
             "description": "Location details",
             "content": {
               "application/json": { "schema": { "$ref": "#/components/schemas/LocationDTO" } }
+            }
+          },
+          "404": {
+            "description": "Location not found with that id",
+            "content": {
+              "application/json": { "schema": { "$ref": "#/components/schemas/ErrorResponse" } }
             }
           }
         }
@@ -288,7 +337,12 @@
               }
             }
           },
-          "404": { "description": "Location not found, or not in any grid" }
+          "404": {
+            "description": "Location not found, or not in any grid",
+            "content": {
+              "application/json": { "schema": { "$ref": "#/components/schemas/ErrorResponse" } }
+            }
+          }
         }
       }
     },
@@ -299,7 +353,15 @@
           { "$ref": "#/components/parameters/LocationIdParam" },
           { "$ref": "#/components/parameters/EntityIdParam" }
         ],
-        "responses": { "200": { "description": "Entity added" } }
+        "responses": {
+          "200": { "description": "Entity added" },
+          "404": {
+            "description": "Location not found with that id",
+            "content": {
+              "application/json": { "schema": { "$ref": "#/components/schemas/ErrorResponse" } }
+            }
+          }
+        }
       },
       "delete": {
         "summary": "Remove entity from location",
@@ -307,7 +369,15 @@
           { "$ref": "#/components/parameters/LocationIdParam" },
           { "$ref": "#/components/parameters/EntityIdParam" }
         ],
-        "responses": { "200": { "description": "Entity removed" } }
+        "responses": {
+          "204": { "description": "Entity removed" },
+          "404": {
+            "description": "Location not found with that id",
+            "content": {
+              "application/json": { "schema": { "$ref": "#/components/schemas/ErrorResponse" } }
+            }
+          }
+        }
       }
     },
     "/api/v1/locations/{locationId}/entity/{entityId}/move": {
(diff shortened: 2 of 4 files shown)


dmccoystephenson:
## Self-review

No human reviewer is configured on this repo, so this is a self-review pass over the full diff. (Posted as a comment rather than a formal review — the review API is not reachable from the environment this ran in.)

Checked for: descriptions that don't match the actual throw site, endpoints given a 404 that can't produce one, endpoints still missing one, `$ref`s that don't resolve, and JSON validity. Verified mechanically — every `4xx`/`5xx` response now carries `content.application/json.schema.$ref -> ErrorResponse`, every `$ref` in the document resolves, and the documented status set per operation was diffed against each controller's `@ResponseStatus` annotations and `throw new *Exception` sites.

### Inline notes

**`src/main/python/preponderous/viron/services/environmentService.py:69`** — this is the actual bug fix (#172). The endpoint returns 204, so the old `return response.status_code == 200` reported failure on every successful delete. `raise_for_status()` on the line above already raises for 4xx/5xx, so anything reaching this line is a 2xx — same shape as `EntityService.delete_entity`.

**`src/test/python/preponderous/viron/services/test_environmentService.py:155`** — this one-character change is what pins the regression: with the old implementation and a 204 mock, `status_code == 200` is `False` and `assert result is True` fails. The test previously mocked 200, encoding the same wrong assumption the implementation made, which is why the suite stayed green over a broken method.

**`docs/openapi/viron-api.json:359`** (`PUT /api/v1/locations/{locationId}/entity/{entityId}`) — the description is deliberately "Location not found with that id" and not "Location or entity not found": `LocationController.addEntityToLocation` only checks that the *location* exists (`LocationController.java:71-73`); a nonexistent entity id falls through to the repository call and surfaces as a 500, not a 404. Same reasoning for the sibling `delete`.

### Two things left as-is rather than silently widened

1. **No `5xx` responses are documented anywhere**, even though `ServiceException` / `EntityServiceException` / the catch-all `Exception` handler all return 500 with the same `ErrorResponse` shape. The new schema description says "every 4xx and 5xx response", which is true of the handler but currently aspirational for the spec. Out of scope for #171 (which enumerated 404s and 400s); a uniform `500` sweep — like the `@Min`/`@Valid` `400` sweep that issue also deferred — is better done in one pass, or dropped entirely once #130 generates the spec from code.

2. **`delete_environment` now always returns `True`.** The `-> bool` return type is arguably vestigial once `raise_for_status()` is the real failure signal, but this matches `EntityService.delete_entity` exactly and keeps the public signature stable for existing SDK callers. Narrowing it to `-> None` would be a breaking change for a cosmetic gain.

No Java source is touched, so CI's `./mvnw test -B` is a regression check only.

---

Issue #138, opened 2026-06-13
EntityFactory is missing the null-ResultSet guard and id==-1 check that EnvironmentFactory has


dmccoystephenson:
## Summary
`EntityFactory` and `EnvironmentFactory` implement the same "get next id from a sequence, then INSERT" pattern, but `EntityFactory` omits two safety checks that `EnvironmentFactory` performs, leaving two confirmed defects on the live entity-creation path (`POST /api/v1/entities/{name}` and `DebugController`).

**1. Unguarded null `ResultSet` → NPE.**
- `src/main/java/preponderous/viron/factories/EntityFactory.java:39-48` — `getNextEntityId()` calls `rs.next()` at line 42 without first checking `rs`:
  ```java
  ResultSet rs = dbInteractions.query("SELECT nextval('viron.entity_id_seq')");
  try {
      if (rs.next()) {            // line 42 — rs may be null
  ```
- `DbInteractions.query(...)` returns `null` on failure (`database/DbInteractions.java:38`), so a failed/locked query throws `NullPointerException` here instead of returning `-1`.
- Contrast `EnvironmentFactory.java:108` (and `:123`, `:138`), which guard it: `if (rs == null) { return -1; }`.

**2. Sentinel `-1` is never checked before the INSERT.**
- `src/main/java/preponderous/viron/factories/EntityFactory.java:28-30`:
  ```java
  int id = getNextEntityId();                  // returns -1 on failure
  String creationDate = new java.util.Date().toString();
  String query = "INSERT INTO viron.entity (entity_id, name, creation_date) VALUES (" + id + ", '" + name + "', '" + creationDate + "')";
  ```
  There is no `if (id == -1)` check, so on failure an INSERT runs with `entity_id = -1`.
- Contrast `EnvironmentFactory.java:32-35`, which does check:
  ```java
  int id = getNextEnvironmentId();
  if (id == -1) {
      log.error("Failed to get next environment id");
      throw new EnvironmentCreationException("Failed to get next environment id");
  }
  ```

## Why it matters
On any DB hiccup, entity creation either crashes with an NPE (case 1) or writes a row with `entity_id = -1` (case 2), risking a primary-key/sequence conflict and a corrupt row — whereas the parallel environment path fails cleanly. The two factories should behave consistently.

## Suggested fix
[...]

PR #174, opened 2026-07-27
refactor: replace ResultSet-returning DbInteractions.query with a RowMapper API

### LINKED ISSUE #144: Replace ResultSet-returning DbInteractions.query with a RowMapper-style API returning List<T>
## Summary
`DbInteractions.query(...)` returns a `ResultSet` (a disconnected `CachedRowSet` as of #143). Because the raw row cursor is handed back to callers, every repository re-implements the same iterate-map-and-error-handle boilerplate. A `RowMapper`-style API that returns `List<T>` would move that loop (and all `SQLException`/null handling) into one place and let repositories return mapped objects directly.

This is a follow-up to #143, which fixed the resource leak (#140) and parameterized the SQL (#139) with minimal blast radius but deliberately left the `ResultSet`-returning shape in place.

## The repeated boilerplate
Each `*RepositoryImpl` query method follows the same pattern — e.g. `src/main/java/preponderous/viron/repositories/EnvironmentRepositoryImpl.java:26-37`:

```java
List<Environment> environments = new ArrayList<>();
ResultSet rs = dbInteractions.query("SELECT * FROM viron.environment");
if (rs == null) {
    log.error("Error finding all environments: ResultSet is null");
    return environments;
}
try {
    while (rs.next()) {
        environments.add(mapResultSetToEnvironment(rs));
    }
} catch (SQLException e) {
    log.error(...);
}
return environments;
```

This `null`-guard + `try`/`while (rs.next())` + `catch (SQLException)` block is repeated across **~18 query methods** in the four repositories (null-guards: `EnvironmentRepositoryImpl` 7, `EntityRepositoryImpl` 6, `GridRepositoryImpl` 4, `LocationRepositoryImpl` 1), and each repository carries a near-identical `mapResultSetTo*` row-mapper helper:
- `EnvironmentRepositoryImpl.java:195` `mapResultSetToEnvironment`
- `EntityRepositoryImpl.java:23` `mapResultSetToEntity`
- `GridRepositoryImpl.java:23` `mapResultSetToGrid`
- `LocationRepositoryImpl.java:23` `mapResultSetToLocation`

## Why it matters
- **Removes duplicated control flow** from ~18 call sites — the iteration, the `null` check, and the `SQLException` swallow all currently live in every repository method.
- **Eliminates a whole class of bug** that this codebase has already hit twice: the missing `if (rs == null)` guard (fixed in #132 for the repositories and #138 for `EntityFactory`). If callers never touch a `ResultSet`, they cannot forget to guard it.
- **Keeps `ResultSet` ownership inside `DbInteractions`**, where it can be opened and closed in one place (reinforcing the #140 fix), rather than handing a cursor to callers.

## Suggested approach
1. Add a functional interface:
   ```java
   @FunctionalInterface
   public interface RowMapper<T> {
       T map(ResultSet rs) throws SQLException;
   }
   ```
2. Add `DbInteractions.query` overloads that own the cursor and return mapped results, e.g.:
   ```java
   <T> List<T> query(String sql, RowMapper<T> mapper, Object... params);          // 0..n rows
   <T> Optional<T> queryOne(String sql, RowMapper<T> mapper, Object... params);    // 0..1 row
   ```
   These iterate, map, and close the `PreparedStatement`/`ResultSet` internally, returning an empty `List`/`Optional` on error.
3. Migrate each repository method to pass its existing `mapResultSetTo*` helper (or a method reference) and return the result directly, deleting the per-method `null`-guard and `try`/`catch` boilerplate.
4. Once all callers are migrated, remove the `ResultSet`-returning `query` (and the `CachedRowSet` machinery it needs).

This is a larger, mechanical refactor touching all four repositories and the factories, so it is intentionally separated from #143. It pairs naturally with #130 (data-layer/contract hygiene).

dmccoystephenson:
## Summary

`DbInteractions.query(...)` returned a raw (disconnected) `ResultSet`, so every repository re-implemented the same iterate-map-and-error-handle boilerplate: a `null` guard, a `while (rs.next())` loop, and a `catch (SQLException)` swallow — repeated across ~18 query methods in the four repositories plus four id-sequence lookups in the factories.

- Add `database/RowMapper.java`, a `@FunctionalInterface` mapping the row a `ResultSet` is currently positioned on.
- Replace `ResultSet query(String, Object...)` with:
  - `<T> List<T> query(String, RowMapper<T>, Object...)`
  - `<T> Optional<T> queryOne(String, RowMapper<T>, Object...)`

  Both own the `PreparedStatement`/`ResultSet`, iterate internally, and return an empty `List`/`Optional` on error. The `CachedRowSet` machinery added for #140 is no longer needed, since nothing escapes the method.
- Migrate all 30 call sites across `EnvironmentRepositoryImpl`, `EntityRepositoryImpl`, `GridRepositoryImpl`, `LocationRepositoryImpl`, `EntityFactory`, and `EnvironmentFactory`. Each query method is now a single expression passing its existing `mapResultSetTo*` helper as a method reference.
- Drop `@Slf4j` from the three repositories whose only logging was the removed error boilerplate.

Because no caller can hold a `ResultSet` any more, the missing-`null`-guard bug this codebase hit twice (#132, #138) is now unrepresentable rather than merely fixed.

No SQL string, endpoint, request/response shape, or DB schema changed — this is internal to the data layer, so the OpenAPI spec and the Python SDK are unaffected.

## Test plan

- [x] Repository tests migrated to the new API via a new test helper, `ResultSetAnswers`, whose Mockito answers drive the `RowMapper` the repository passed over a mocked `ResultSet`. This keeps each repository's row mapping covered by its own tests even though the cursor is no longer visible to the repository.
- [x] Per-repository `ResultSet is null` / `SQLException thrown` cases collapse into one `...WhenQueryFails` case each, since that handling now lives in one place. The underlying behaviour is covered directly in `DbInteractionsTest` against H2, which gains cases for: `query`/`queryOne` on invalid SQL, no matching rows, multiple matching rows, and a `RowMapper` that throws mid-iteration.
- [x] `EntityFactoryTest`'s two #138 regression cases (null cursor, no row) become the single empty-`Optional` case they now both produce.
- [ ] `./mvnw test -B` — **not runnable in this environment** (no Maven access); relying on the CI workflow's `./mvnw compile -B` + `./mvnw test -B` on this branch as the authoritative signal.
- [x] `pytest` — not applicable, no Python SDK changes.

Closes #144

---

### DIFF
--- a/src/main/java/preponderous/viron/database/DbInteractions.java
+++ b/src/main/java/preponderous/viron/database/DbInteractions.java
@@ -9,7 +9,6 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 
@@ -59,7 +58,8 @@ public class DbInteractions {
             }
         } catch (SQLException e) {
             log.error("Error executing query: {}", e.getMessage());
-            return Collections.emptyList();
+            // Discard any partially mapped rows rather than reporting a truncated result.
+            return new ArrayList<>();
         }
         return results;
     }
--- a/src/main/java/preponderous/viron/database/DbInteractions.java
+++ b/src/main/java/preponderous/viron/database/DbInteractions.java
@@ -8,9 +8,10 @@ import java.sql.DriverManager;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
-
-import javax.sql.rowset.CachedRowSet;
-import javax.sql.rowset.RowSetProvider;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
 
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -34,30 +35,59 @@ public class DbInteractions {
     }
 
     /**
-     * Execute a parameterized SELECT and return a disconnected result.
+     * Execute a parameterized SELECT and map every returned row.
      *
      * <p>The query is run through a {@link PreparedStatement} so caller-supplied
      * values are bound as parameters rather than concatenated into SQL. The
-     * statement and live result set are closed before returning; the rows are
-     * copied into a {@link CachedRowSet} so callers can read them without leaking
-     * JDBC resources.
+     * statement and result set are owned here and closed before returning, so no
+     * JDBC resource ever reaches the caller.
      *
      * @param query  SQL with {@code ?} placeholders for each parameter
+     * @param mapper maps each row to a result object
      * @param params values to bind to the placeholders, in order
-     * @return a disconnected {@link ResultSet} of the rows, or {@code null} on error
+     * @param <T>    mapped result type
+     * @return the mapped rows in result-set order, or an empty list if the query failed
      */
-    public ResultSet query(String query, Object... params) {
+    public <T> List<T> query(String query, RowMapper<T> mapper, Object... params) {
+        List<T> results = new ArrayList<>();
         try (PreparedStatement statement = connection.prepareStatement(query)) {
             bindParameters(statement, params);
             try (ResultSet rs = statement.executeQuery()) {
-                CachedRowSet cachedRowSet = RowSetProvider.newFactory().createCachedRowSet();
-                cachedRowSet.populate(rs);
-                return cachedRowSet;
+                while (rs.next()) {
+                    results.add(mapper.map(rs));
+                }
+            }
+        } catch (SQLException e) {
+            log.error("Error executing query: {}", e.getMessage());
+            return Collections.emptyList();
+        }
+        return results;
+    }
+
+    /**
+     * Execute a parameterized SELECT and map its first row, if any.
+     *
+     * <p>Behaves like {@link #query(String, RowMapper, Object...)} but stops after the
+     * first row; any further rows the query happens to return are ignored.
+     *
+     * @param query  SQL with {@code ?} placeholders for each parameter
+     * @param mapper maps the first row to a result object
+     * @param params values to bind to the placeholders, in order
+     * @param <T>    mapped result type
+     * @return the mapped first row, or an empty {@link Optional} if there were no rows or the query failed
+     */
+    public <T> Optional<T> queryOne(String query, RowMapper<T> mapper, Object... params) {
+        try (PreparedStatement statement = connection.prepareStatement(query)) {
+            bindParameters(statement, params);
+            try (ResultSet rs = statement.executeQuery()) {
+                if (rs.next()) {
+                    return Optional.ofNullable(mapper.map(rs));
+                }
             }
         } catch (SQLException e) {
             log.error("Error executing query: {}", e.getMessage());
         }
-        return null;
+        return Optional.empty();
     }
 
     /**
@@ -106,5 +136,5 @@ public class DbInteractions {
         }
         return connection;
     }
-    
+
 }
--- a/src/main/java/preponderous/viron/database/RowMapper.java
+++ b/src/main/java/preponderous/viron/database/RowMapper.java
@@ -1,0 +1,29 @@
+// Copyright (c) 2024 Preponderous Software
+// MIT License
+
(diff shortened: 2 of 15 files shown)


dmccoystephenson:
## Self-review

Reviewed the full diff against this repo's data-layer conventions (parameterized SQL, JDBC resource ownership, one exception type per failure mode, tests on every public method). **CI is green** on this branch (`./mvnw compile -B` + `./mvnw test -B`), which is the authoritative signal here — Maven isn't runnable in the environment this cycle ran in.

*(Posted as a comment rather than a formal Review object: this dispatch has no access to the reviews API. Inline findings are folded in below as `path:line — finding`.)*

### Findings

**1. `src/main/java/preponderous/viron/database/DbInteractions.java:64` — inconsistent list mutability on the error path.** *(fixed in 6985fea)*

The success path returned a mutable `ArrayList`, but the error path returned `Collections.emptyList()`, which is immutable. A caller that mutates the result would then work fine normally and throw `UnsupportedOperationException` only when the query happened to fail — exactly the kind of path that gets exercised first in production. No current caller mutates a repository result (checked `services/` and `controllers/`), so this was latent rather than live, but the asymmetry is worth not shipping. Both paths now return `new ArrayList<>()`.

**2. `src/main/java/preponderous/viron/database/DbInteractions.java:52` — partial results are discarded when a mapper throws mid-iteration.** *(intentional, documented)*

This is a deliberate behaviour change from the old code, which returned whatever rows had accumulated before the `SQLException`. Returning a silently truncated list is worse than returning none: a caller can't distinguish "3 rows" from "3 of 40 rows". The Javadoc states it, and `DbInteractionsTest.query_whenMapperThrows_returnsEmptyListAndDoesNotThrow` pins it.

**3. `src/main/java/preponderous/viron/database/DbInteractions.java:88` — `queryOne` collapses "mapper returned null" into `Optional.empty()`.**

`Optional.ofNullable(mapper.map(rs))` means a mapper that maps a row to `null` is indistinguishable from no row at all. Every mapper in the codebase returns a non-null object or an `int`, so this is unreachable today, and `Optional.of` would trade an unreachable ambiguity for a reachable NPE. Leaving as-is deliberately.

**4. Scope check — SQL strings deliberately untouched.**

Several migrated call sites still interpolate ints into SQL (`"... WHERE grid_id = " + id`) rather than binding them, even though the new signature takes varargs params right there. That's not an injection vector (the values are `int`), but it's a leftover from #139's partial parameterization and it now reads oddly next to a params-accepting API. I kept this PR to the `RowMapper` migration alone so the diff has exactly one idea in it; filed as a follow-up rather than smuggled in.

### Checks that came back clean

- **No unparameterized user input / unclosed JDBC resources.** `PreparedStatement` and `ResultSet` are both in try-with-resources inside `DbInteractions`, and neither escapes the method any more — which is the structural point of #144. No `String`-typed value is concatenated into any SQL.
- **No new exception types**, so no `GlobalExceptionHandler` gap. Failure handling stays "log and return empty", as before.
- **No DTO, endpoint, request/response shape, or schema change** — so no OpenAPI spec drift and no Python SDK drift. Verified `docs/openapi/viron-api.json` still matches all 29 controller mappings, and `src/main/python/.../services/*.py` is unaffected.
- **Test coverage on new public methods.** `query`/`queryOne` are covered in `DbInteractionsTest` against H2 (invalid SQL, no rows, one row, multiple rows, mapper throwing). `RowMapper` is an interface with no implementation to test directly; it's exercised through every repository test.
- **Mapper coverage was not lost in the migration.** This was the main risk: stubbing the new API with a canned `List` would have left every `mapResultSetTo*` helper untested. `ResultSetAnswers` avoids that by driving the `RowMapper` the repository actually passed over a mocked `ResultSet`, so each repository still asserts its own column-to-field mapping.
- **Only genuinely redundant tests were dropped.** The removed per-repository "ResultSet is null" and "SQLException thrown" cases tested error handling that no longer exists in the repositories; the behaviour they pinned is now tested once, for real, in `DbInteractionsTest`. Every repository method keeps a rows-exist case, an empty case, and a query-fails case. `EntityFactoryTest`'s two #138 regression cases became the one empty-`Optional` case they both now produce.
- **`@Slf4j` removals are safe.** Dropped only from the four repositories whose sole logging was the deleted boilerplate; both factories keep it and still log.

---

PR #179, opened 2026-07-30
test: add CI guard for OpenAPI spec/controller route drift

dmccoystephenson:
## Summary
- Adds `OpenApiSpecDriftTest`, a full-context Spring Boot test that fetches the live route list from springdoc's `/v3/api-docs` (with `viron.debug.enabled=true` so `DebugController`'s routes are included) and asserts it matches the route set (verb + path) declared in `docs/openapi/viron-api.json`.
- Because this runs under the existing `mvn test` step, CI now fails automatically if a controller's routes drift from the checked-in OpenAPI contract — no separate CI workflow step was needed.
- This addresses the **CI spec-drift guard** half of #130. The other half of that issue (wiring `openapi-generator-maven-plugin` for codegen, or generating the spec from code) is a materially larger, separate architectural change and is left open on #130 — this PR does not claim to close it.

## Verification
- Confirmed the guard actually detects drift: temporarily removed `POST /api/v1/environments` from the static spec and re-ran the test — it failed with a clear diff of the mismatched route sets. Reverted before committing.
- Confirmed it passes against the current (non-drifted) spec and controllers.

## Test plan
- [x] `mvn test -B` — full suite green (265 tests, 0 failures)
- [x] `mvn test -B -Dtest=OpenApiSpecDriftTest` — passes against current spec
- [x] Manually broke the static spec and confirmed the test fails with a useful message, then reverted

---

### DIFF
--- a/src/test/java/preponderous/viron/OpenApiSpecDriftTest.java
+++ b/src/test/java/preponderous/viron/OpenApiSpecDriftTest.java
@@ -1,0 +1,73 @@
+// Copyright (c) 2024 Preponderous Software
+// MIT License
+
+package preponderous.viron;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.security.test.context.support.WithMockUser;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.test.web.servlet.MockMvc;
+
+import java.io.File;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+/**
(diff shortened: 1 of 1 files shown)


dmccoystephenson:
Self-review rubric:
- Scope: PASS — single new test file (OpenApiSpecDriftTest.java), directly implements the CI spec-drift-guard half of #130; no unrelated changes.
- Tests-new: PASS — this PR's payload *is* a test; verified it actually detects drift by temporarily deleting `POST /api/v1/environments` from docs/openapi/viron-api.json and re-running `mvn test -Dtest=OpenApiSpecDriftTest` (failed with a clear route-set diff), then reverted before committing.
- Tests-fix: N/A — not a bug fix, no regression to characterize.
- Sibling structure: PASS — matches the existing `@SpringBootTest` / `@AutoConfigureMockMvc` / `@WithMockUser` / `@DirtiesContext` / `@TestPropertySource(viron.debug.enabled=true)` pattern used by DebugControllerTest and sibling controller tests; placed alongside VironApplicationTest as the other top-level app-level test.
- Sibling renames: N/A — no renames.
- Docs: PASS — no controller/spec behavior changed, so no doc source needed updating. docs/REBUILD_PLAN.md's codegen step (the other half of #130) is untouched and still accurately describes outstanding work.
- Issue resolution: PASS (partial, stated as such) — this only resolves the "CI step that fails on spec/implementation drift" half of #130; the openapi-generator codegen half remains open, so the PR does not include a `Closes #130` and the PR body says so explicitly.
- CI: PASS — green on the PR head SHA (build check, run 30530882416).

Repo-specific:
- DTO boundary: N/A — no controller changes.
- Spec alignment: N/A — no endpoint changes; this PR only guards existing alignment.
- Java/Python parallelism: N/A — pure Java-side test tooling, no model/service behavior changed.
- Override correctness: N/A — no @Override added.

No FAILs found. Full suite: 265/265 passing locally (mvn test -B) and in CI.

---

PR #186, opened 2026-08-02
fix: stop rewrapping domain exceptions as generic ServiceException; add LocationService tests

### LINKED ISSUE #183: LocationService rewraps its own NotFoundException/ConflictException as a generic ServiceException
Several of `LocationService`'s failure branches throw a domain exception **from inside the `try` block**, where the method's own trailing `catch (Exception e)` immediately catches and rewraps it.

Because `NotFoundException` and `ConflictException` both extend `ServiceException` (which extends `RuntimeException`), they are caught by `catch (Exception e)` and rethrown as a plain `ServiceException`. `GlobalExceptionHandler` then maps that to **500 Internal Server Error** instead of the intended **404** / **409**.

## Affected methods

`src/main/java/preponderous/viron/services/LocationService.java`:

| Method | Thrown inside `try` | Intended status | Actual status |
|---|---|---|---|
| `getLocationById` | `NotFoundException` | 404 | 500 |
| `getLocationOfEntity` | `NotFoundException` | 404 | 500 |
| `addEntityToLocation` | `NotFoundException`, `ServiceException` | 404 / 500 | 500 (message replaced) |
| `removeEntityFromLocation` | `NotFoundException`, `ServiceException` | 404 / 500 | 500 (message replaced) |
| `removeEntityFromCurrentLocation` | `NotFoundException`, `ServiceException` | 404 / 500 | 500 (message replaced) |
| `isLocationOccupied` | `ServiceException` | 500 | 500 (message replaced) |
| `moveEntityToLocation` | `NotFoundException`, `ConflictException` | 404 / 409 | 500 |

The `catch (HttpClientErrorException.NotFound e)` / `catch (HttpClientErrorException.Conflict e)` clauses are **not** affected — they sit outside the `try`, so what they throw propagates correctly.

## Reachability

`RestTemplate`'s default error handler throws on 4xx/5xx, so the explicit `response.getStatusCode() == HttpStatus.NOT_FOUND` branches are largely defensive. The reachable case today is a **2xx response with a null body**: `getLocationById` and `getLocationOfEntity` both check `response.getBody() != null` and throw `NotFoundException` when it is null — that path surfaces as 500 rather than 404.

## Proposed fix

Add a `catch (ServiceException e) { throw e; }` clause ahead of each `catch (Exception e)` so already-domain-typed exceptions pass through unchanged.

## Related

`EntityService.createEntity` has the same shape (line 145 throws `EntityServiceException` inside the `try`, caught at line 150), but since it is rewrapped as the same type it only loses the specific message — no status-code change. Worth fixing alongside, but lower priority.

---

---

### LINKED ISSUE #184: Add unit tests for LocationService
`src/main/java/preponderous/viron/services/LocationService.java` is the largest service in the repo (248 lines, 11 public methods) and has no test class at all — `src/test/java/preponderous/viron/services/` contains only `EntityServiceTest` and `GridServiceTest` (the latter added in PR #182).

Untested public methods:

- `getAllLocations()`
- `getLocationById(int)`
- `getLocationsInEnvironment(int)`
- `getLocationsInGrid(int)`
- `getLocationOfEntity(int)`
- `addEntityToLocation(int, int)`
- `removeEntityFromLocation(int, int)`
- `removeEntityFromCurrentLocation(int)`
- `getEntityIdsAtLocation(int)`
- `isLocationOccupied(int)`
- `moveEntityToLocation(int, int)`

## Proposed work

Add `src/test/java/preponderous/viron/services/LocationServiceTest.java` following the plain-JUnit + Mockito pattern established by `GridServiceTest`: mock `RestTemplateBuilder`/`RestTemplate`, stub `additionalInterceptors(...)` to return the builder, construct the service with a `ServiceConfig` pointed at `http://localhost:8080`, and assert per method:

- success path returns the mapped body
- `HttpClientErrorException.NotFound` maps to `NotFoundException` where the service declares it
- `HttpClientErrorException.Conflict` maps to `ConflictException` in `moveEntityToLocation`
- an arbitrary `RestClientException` is wrapped in `ServiceException` with the method's message
- null/empty body maps to `Collections.emptyList()` on the list-returning methods

Note `LocationService` builds its `RestTemplate` once in the constructor (unlike `GridService`, which calls `restTemplateBuilder.build()` per request), so the mock builder must return the mock template from that single constructor-time `build()` call.

This should be done together with (or after) the exception-rewrap fix in #183, since the intended-vs-actual exception types differ until that is fixed.

---

dmccoystephenson:
## Summary

- **Fix (#183)** — `LocationService` threw `NotFoundException` / `ConflictException` **from inside** its `try` blocks, where the method's own trailing `catch (Exception e)` immediately caught and rewrapped them. Because both extend `ServiceException` (→ `RuntimeException`), `GlobalExceptionHandler` mapped the result to **500** instead of the intended **404** / **409**. Added a `catch (ServiceException e) { throw e; }` passthrough ahead of each generic catch in the 7 affected methods: `getLocationById`, `getLocationOfEntity`, `addEntityToLocation`, `removeEntityFromLocation`, `removeEntityFromCurrentLocation`, `isLocationOccupied`, `moveEntityToLocation`.
- Applied the same passthrough to `EntityService.createEntity`, called out under "Related" in #183 — it rewrapped as the same type, so no status change, but the specific `"Created entity response was null"` message was being replaced by the generic `"Error creating entity"`. `EntityServiceTest`'s null-body expectation is updated to the now-preserved message.
- **Tests (#184)** — new `src/test/java/preponderous/viron/services/LocationServiceTest.java` covering all 11 public methods, following the plain-JUnit + Mockito pattern from `GridServiceTest` (mock `RestTemplateBuilder`, stub `additionalInterceptors(...)` to return the builder, single constructor-time `build()`).

Per method the suite asserts, as applicable: success path, null body → `Collections.emptyList()` / `NotFoundException` / `ServiceException`, non-2xx status branch, `HttpClientErrorException.NotFound` → `NotFoundException`, `HttpClientErrorException.Conflict` → `ConflictException`, and arbitrary `RestClientException` → `ServiceException`. The `ServiceException` assertions use `isExactlyInstanceOf` so they fail if a `NotFoundException`/`ConflictException` is silently downgraded — these are the regression guards for #183.

No endpoint, request/response shape, or auth surface changed, so `docs/openapi/viron-api.json` and the Python SDK need no update.

## Test plan

- [ ] `./mvnw compile -B` — not run locally: this sandbox has no usable Maven, so CI's compile step is the authoritative signal.
- [ ] `./mvnw test -B` — same; verified via the CI run on this branch before merge.
- [x] `pytest` — n/a, `src/main/python` / `src/test/python` untouched.

Closes #183
Closes #184

---

### DIFF
--- a/src/main/java/preponderous/viron/services/EntityService.java
+++ b/src/main/java/preponderous/viron/services/EntityService.java
@@ -9,6 +9,7 @@ import preponderous.viron.config.ServiceConfig;
 import preponderous.viron.dto.CreateEntityRequest;
 import preponderous.viron.dto.UpdateEntityNameRequest;
 import preponderous.viron.exceptions.EntityServiceException;
+import preponderous.viron.exceptions.ServiceException;
 import preponderous.viron.models.Entity;
 
 import java.util.Arrays;
@@ -147,6 +148,8 @@ public class EntityService {
 
             return response.getBody();
 
+        } catch (ServiceException e) {
+            throw e;
         } catch (Exception e) {
             log.error("Failed to create entity with name {}: {}", name, e.getMessage());
             throw new EntityServiceException("Error creating entity", e);
--- a/src/main/java/preponderous/viron/services/LocationService.java
+++ b/src/main/java/preponderous/viron/services/LocationService.java
@@ -53,6 +53,8 @@ public class LocationService {
             throw new NotFoundException("Location not found with id: " + id);
         } catch (HttpClientErrorException.NotFound e) {
             throw new NotFoundException("Location not found with id: " + id);
+        } catch (ServiceException e) {
+            throw e;
         } catch (Exception e) {
             log.error("Error getting location by id {}: {}", id, e.getMessage());
             throw new ServiceException("Error getting location by id: " + id, e);
@@ -104,6 +106,8 @@ public class LocationService {
             throw new NotFoundException("Location not found for entity: " + entityId);
         } catch (HttpClientErrorException.NotFound e) {
             throw new NotFoundException("Location not found for entity: " + entityId);
+        } catch (ServiceException e) {
+            throw e;
         } catch (Exception e) {
             log.error("Error getting location of entity {}: {}", entityId, e.getMessage());
             throw new ServiceException("Error getting location of entity: " + entityId, e);
@@ -128,6 +132,8 @@ public class LocationService {
             }
         } catch (HttpClientErrorException.NotFound e) {
             throw new NotFoundException("Location or entity not found");
+        } catch (ServiceException e) {
+            throw e;
         } catch (Exception e) {
             log.error("Error adding entity {} to location {}: {}", entityId, locationId, e.getMessage());
             throw new ServiceException("Error adding entity to location", e);
@@ -152,6 +158,8 @@ public class LocationService {
             }
         } catch (HttpClientErrorException.NotFound e) {
             throw new NotFoundException("Location or entity not found");
+        } catch (ServiceException e) {
+            throw e;
         } catch (Exception e) {
             log.error("Error removing entity {} from location {}: {}", entityId, locationId, e.getMessage());
             throw new ServiceException("Error removing entity from location", e);
@@ -175,6 +183,8 @@ public class LocationService {
             }
         } catch (HttpClientErrorException.NotFound e) {
             throw new NotFoundException("Entity not found");
+        } catch (ServiceException e) {
+            throw e;
         } catch (Exception e) {
             log.error("Error removing entity {} from current location: {}", entityId, e.getMessage());
             throw new ServiceException("Error removing entity from current location", e);
@@ -212,6 +222,8 @@ public class LocationService {
             throw new ServiceException("Failed to check occupancy for location: " + locationId);
         } catch (HttpClientErrorException.NotFound e) {
             throw new NotFoundException("Location not found with id: " + locationId);
+        } catch (ServiceException e) {
+            throw e;
         } catch (Exception e) {
             log.error("Error checking occupancy for location {}: {}", locationId, e.getMessage());
             throw new ServiceException("Error checking occupancy for location: " + locationId, e);
@@ -241,6 +253,8 @@ public class LocationService {
             throw new NotFoundException("Entity or location not found");
         } catch (HttpClientErrorException.Conflict e) {
             throw new ConflictException("Target location " + locationId + " is already occupied");
+        } catch (ServiceException e) {
+            throw e;
         } catch (Exception e) {
             log.error("Error moving entity {} to location {}: {}", entityId, locationId, e.getMessage());
             throw new ServiceException("Error moving entity to location", e);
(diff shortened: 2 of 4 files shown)


dmccoystephenson:
## Self-review

Reviewed the full diff against the repo's service/exception conventions. **No blocking issues found.** CI (`./mvnw compile -B` + `./mvnw test -B`) is green on this branch — that is the authoritative signal here, since this sandbox has no usable local Maven.

Verified:
- All 7 methods listed in #183's table got the passthrough, and the fix is placed *after* the `HttpClientErrorException.NotFound`/`.Conflict` catches — those types are unrelated to `ServiceException`, so ordering is both compile-legal and behaviourally irrelevant.
- No new exception type was introduced, so `GlobalExceptionHandler` needs no new handler method. The existing `NotFoundException` → 404 / `ConflictException` → 409 / `ServiceException` → 500 mappings are what the fix now actually reaches.
- No endpoint, request/response shape, or `permitAll()` matcher changed → `docs/openapi/viron-api.json`, `SecurityConfigTest`, and the Python SDK under `src/main/python` correctly need no update.
- `DebugController` is the only consumer of these services and its tests mock them, so no controller test depends on the old 500 behaviour.
- No SQL / JDBC touched, so the unparameterized-SQL and unclosed-resource checks are n/a for this diff.

### Notes (non-blocking)

`src/main/java/preponderous/viron/services/LocationService.java:194` — `getEntityIdsAtLocation` is now the only method in this class *without* the `catch (ServiceException e) { throw e; }` passthrough. That is correct today: it throws nothing domain-typed inside its `try` (the `NotFoundException` comes from the `HttpClientErrorException.NotFound` catch, which sits outside). Deliberately left alone to avoid dead code, but it means the class is no longer uniform — if someone later adds a `throw new NotFoundException(...)` inside that `try`, #183 silently returns for this one method.

`src/main/java/preponderous/viron/services/EntityService.java:151` — this changes the user-visible error body for the null-response-body case from `"Error creating entity"` to `"Created entity response was null"`. The HTTP status is unchanged (500 either way, via the `EntityServiceException` handler), so this is a message-fidelity improvement rather than a contract change, but it is a behavioural change beyond the strict letter of #183's title. It is what #183's "Related" section asks for; `EntityServiceTest`'s expectation is updated in the same commit.

`src/test/java/preponderous/viron/services/LocationServiceTest.java:246` (and the sibling `*_NotFoundStatus_*` / `*_ConflictStatus_*` / `*_NonSuccessStatus_*` cases) — these stub a `ResponseEntity` carrying a 4xx status, which `RestTemplate`'s default error handler makes unreachable in production; it throws `HttpClientErrorException` before the service ever inspects `response.getStatusCode()`. They are worth keeping — they are precisely the regression guards for the rewrapping bug, and they pin the defensive branches — but the tests that prove real-world 404/409 behaviour are the `*_HttpNotFound_*` / `*_HttpConflict_*` ones. Flagging so the coverage is not read as stronger than it is.

`src/test/java/preponderous/viron/services/LocationServiceTest.java:53` — the `nullRequestEntity()` matcher helper exists because every `exchange(...)` call in `LocationService` passes a `null` request entity. Consequence: nothing in this suite asserts that `AuthTokenInterceptor` is actually wired onto the built `RestTemplate`. That is the same gap `GridServiceTest` and `EntityServiceTest` have, so it is consistent with repo convention rather than a regression — but no service test in this repo currently covers the auth-header path.

---

PR #187, opened 2026-08-04
feat: add EnvironmentService tests and align its error handling

### LINKED ISSUE #185: Add unit tests for EnvironmentService
`src/main/java/preponderous/viron/services/EnvironmentService.java` has no test class. `src/test/java/preponderous/viron/services/` contains only `EntityServiceTest` and `GridServiceTest` (the latter added in PR #182); `LocationService` is tracked separately.

Untested public methods:

- `getAllEnvironments()`
- `getEnvironmentById(int)`
- `getEnvironmentByName(String)`
- `getEnvironmentOfEntity(int)`
- `createEnvironment(String, int, int)` — square-grid overload
- `createEnvironment(String, int, int, int)` — rows/columns overload
- `deleteEnvironment(int)`
- `updateEnvironmentName(int, String)`

## Proposed work

Add `src/test/java/preponderous/viron/services/EnvironmentServiceTest.java` following the plain-JUnit + Mockito pattern established by `GridServiceTest`.

Points worth covering specifically:

- The two `createEnvironment` overloads build different `CreateEnvironmentRequest` payloads (`gridSize` vs. `numRows`/`numColumns`). Capture the `HttpEntity` argument and assert the request body actually differs — nothing currently pins that contract.
- `updateEnvironmentName` sends a `UpdateEnvironmentNameRequest` body over `PATCH`; assert both method and body.
- `deleteEnvironment`/`updateEnvironmentName` return `response.getStatusCode().is2xxSuccessful()`.

## Secondary finding

Unlike `GridService`, `LocationService`, and `EntityService`, `EnvironmentService` does not wrap `RestTemplate` failures at all — it only checks `response.getStatusCode().isError()`, which is unreachable under `RestTemplate`'s default error handler (4xx/5xx throw `HttpClientErrorException`/`HttpServerErrorException` before the check runs). The practical effect is that a failed call propagates a raw `RestClientException` to the caller instead of the repo's `ServiceException`, which `GlobalExceptionHandler` then falls through to its generic `Exception` handler ("An unexpected error occurred") rather than the `ServiceException` handler.

Aligning `EnvironmentService` with the try/catch-and-wrap style of its sibling services would be a reasonable part of this work, but is called out separately here so it can be split off if preferred.

---

dmccoystephenson:
## Summary

- `EnvironmentServiceTest` is added, covering all eight public methods of `EnvironmentService` — the only client service in the repo that had no test class. Success paths, null-body and non-2xx branches, `NotFound` mapping, and generic `RestClientException` wrapping are all exercised, following the plain-JUnit + Mockito pattern established by `GridServiceTest` and `LocationServiceTest`.
- The request bodies of the two `createEnvironment` overloads are captured and asserted, so the `gridSize` shorthand and the `numRows`/`numColumns` form are pinned as genuinely different payloads; the `PATCH` rename payload is pinned the same way.
- `EnvironmentService`'s error handling is aligned with the try/catch-and-wrap style of its siblings. The previous `response.getStatusCode().isError()` checks were unreachable under `RestTemplate`'s default error handler, so a failed call propagated a raw `RestClientException` and was mapped by `GlobalExceptionHandler` to its generic "An unexpected error occurred" branch rather than the `ServiceException` branch.
- `HttpClientErrorException.NotFound` and 404 responses are now translated to `NotFoundException` for the by-id, by-name, of-entity, delete, and rename calls, so a missing environment surfaces as 404 instead of 500. A `catch (ServiceException e) { throw e; }` passthrough sits ahead of each generic catch, matching the fix made in #186 so domain exceptions are not rewrapped.
- A null or non-OK list response from `getAllEnvironments()` now yields an empty list instead of a `NullPointerException` out of `Arrays.asList(null)`.

No endpoint URL, request shape, or response shape was changed, so `docs/openapi/viron-api.json` and the Python SDK under `src/main/python` require no update.

## Test plan

- [ ] `./mvnw test -B` — not runnable in the sandbox this branch was authored in (no local JDK/Maven); the CI workflow's `./mvnw test -B` step on this branch is the authoritative signal and is checked before merge.
- [x] `pytest` — not applicable; no Python SDK file was touched.

Closes #185

### DIFF
--- a/src/main/java/preponderous/viron/services/EnvironmentService.java
+++ b/src/main/java/preponderous/viron/services/EnvironmentService.java
@@ -4,24 +4,30 @@
 package preponderous.viron.services;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.web.client.RestTemplateBuilder;
 import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Service;
+import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestTemplate;
 
+import lombok.extern.slf4j.Slf4j;
 import preponderous.viron.config.AuthTokenInterceptor;
 import preponderous.viron.config.ServiceConfig;
 import preponderous.viron.dto.CreateEnvironmentRequest;
 import preponderous.viron.dto.UpdateEnvironmentNameRequest;
+import preponderous.viron.exceptions.NotFoundException;
 import preponderous.viron.exceptions.ServiceException;
 import preponderous.viron.models.Environment;
 
 @Service
+@Slf4j
 public class EnvironmentService {
     private final RestTemplateBuilder restTemplateBuilder;
     private final ServiceConfig serviceConfig;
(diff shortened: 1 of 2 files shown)


dmccoystephenson:
## Self-review

The full diff was read and checked against this repo's conventions: sibling-service error-handling style, `GlobalExceptionHandler` coverage, DTO/OpenAPI/Python-SDK drift, and test coverage of every public method. No blocking issue was found. Five observations are recorded below; none require a change, and the reasoning for leaving each as-is is given.

**Findings**

`src/main/java/preponderous/viron/services/EnvironmentService.java:47` — The list path treats anything other than `HttpStatus.OK` as an empty result, so a hypothetical `204`/`206` would be reported as zero environments rather than as an error. This is deliberate: it mirrors `LocationService.getAllLocations()` exactly, and consistency across the four client services was judged more valuable than a stricter check here. The old `Arrays.asList(response.getBody())` would have thrown a `NullPointerException` on a null body, so this is strictly an improvement.

`src/main/java/preponderous/viron/services/EnvironmentService.java:60` — For the single-resource lookups, the only way the trailing `throw new NotFoundException(...)` is reached is a non-error status carrying a null body, because 4xx/5xx are raised by `RestTemplate`'s default error handler before the check runs. Labelling that case "not found" is the same choice `LocationService.getLocationById()` makes, and is the reason the previously unreachable `getStatusCode().isError()` branch was replaced rather than kept.

`src/main/java/preponderous/viron/services/EnvironmentService.java:64` — The `NotFoundException` raised from the `HttpClientErrorException.NotFound` catch does not carry the original exception as its cause. This matches `LocationService`, where the cause is likewise dropped for the 404 translation; the cause is preserved on every generic `ServiceException` wrap. Left consistent with the sibling rather than diverging in one file.

`src/main/java/preponderous/viron/services/EnvironmentService.java:160` — `deleteEnvironment` and `updateEnvironmentName` now return a literal `true` instead of `response.getStatusCode().is2xxSuccessful()`. The two are equivalent, because every non-2xx status is thrown on the lines immediately above; the literal makes the post-condition explicit.

`src/main/java/preponderous/viron/services/EnvironmentService.java:60` (contract change) — `getEnvironmentById`, `getEnvironmentByName`, and `getEnvironmentOfEntity` now raise `NotFoundException` where a null body previously produced a `null` return. `DebugController` is the only in-repo consumer of this service and calls `createEnvironment` alone, so no call site is affected. `NotFoundException` is already wired into `GlobalExceptionHandler` (404), and the `catch (ServiceException e) { throw e; }` passthrough ahead of each generic catch keeps it from being rewrapped — the same regression fixed in #186.

**Checked and clear**

- Every new failure mode maps to an exception type already handled in `GlobalExceptionHandler`; no new exception type was introduced.
- No repository, SQL, or JDBC resource handling is touched by this diff.
- No DTO was added or changed, so no validation or `@Schema` annotation is missing.
- No endpoint path, request shape, or response shape changed, so `docs/openapi/viron-api.json` and the Python SDK are not in drift; the existing `OpenApiSpecDriftTest` continues to guard that.
- All eight public methods of `EnvironmentService` are covered by the new test class, including the previously unpinned distinction between the two `createEnvironment` overloads' request bodies.

Local verification was not possible in the sandbox this branch was authored in (no JDK or Maven available), so the CI workflow's `./mvnw test -B` run is treated as the authoritative signal and is checked before any merge.

This review comment was drafted during a Gardener session (https://github.com/Stephenson-Software/gardener).

PR #189, opened 2026-08-07
feat: add missing Java client service methods for three implemented endpoints

### LINKED ISSUE #188: Java client services are missing three implemented endpoints that the Python SDK already exposes
## Summary

Three endpoints that are implemented by the controllers and already reachable from the Python SDK have no counterpart in the Java client services under `src/main/java/preponderous/viron/services/`. A Java consumer of Viron therefore has to fall back to a raw `RestTemplate` call for functionality the Python consumer gets from the SDK.

## Verified gaps

| Endpoint | Controller | Python SDK | Java client service |
|---|---|---|---|
| `PATCH /api/v1/grids/{id}/name` | `GridController.updateGridName` (`GridController.java:54`) | `gridService.py:44` `update_grid_name` | absent from `GridService` |
| `GET /api/v1/locations/grid/{gridId}/unoccupied` | `LocationController.getUnoccupiedLocationsInGrid` (`LocationController.java:56`) | `locationService.py:41` `get_unoccupied_locations_in_grid` | absent from `LocationService` |
| `GET /api/v1/locations/{locationId}/neighbors` | `LocationController.getNeighbors` (`LocationController.java:117`) | `locationService.py:46` `get_neighbors` | absent from `LocationService` |

`GridService` currently exposes only `getAllGrids`, `getGridById`, `getGridsInEnvironment`, and `getGridOfEntity`. `LocationService` exposes eleven methods, none of which cover the unoccupied-in-grid or neighbors routes. `EntityService` and `EnvironmentService` were checked the same way and have full parity with their controllers, so the gap is confined to these three routes.

## Proposed work

- `GridService.updateGridName(int id, String name)` — `PATCH` with an `UpdateGridNameRequest` body, returning `boolean`, matching the shape of `EnvironmentService.updateEnvironmentName`.
- `LocationService.getUnoccupiedLocationsInGrid(int gridId)` and `LocationService.getNeighbors(int locationId)` — `GET` returning `List<Location>`, following the existing `getLocationsInGrid` pattern including its `catch (ServiceException e) { throw e; }` passthrough and `NotFoundException` translation.
- Matching tests in `GridServiceTest` and `LocationServiceTest`, covering the success path, the `NotFound` mapping, and the generic `RestClientException` wrapping, as the sibling methods there already do.

No controller, DTO, OpenAPI spec, or Python SDK change is required; only the Java client services and their tests are affected.

dmccoystephenson:
## Summary

- `GridService.updateGridName(int id, String name)` was added, issuing `PATCH /api/v1/grids/{id}/name` with an `UpdateGridNameRequest` body and mirroring the shape and error handling of `EnvironmentService.updateEnvironmentName` (404 mapped to `NotFoundException`, non-2xx and transport errors wrapped in `ServiceException`).
- `LocationService.getUnoccupiedLocationsInGrid(int gridId)` was added, issuing `GET /api/v1/locations/grid/{gridId}/unoccupied` and following the existing `getLocationsInGrid` pattern (empty list on non-OK or null body, `ServiceException` wrapping).
- `LocationService.getNeighbors(int locationId)` was added, issuing `GET /api/v1/locations/{locationId}/neighbors` and following the existing `getEntityIdsAtLocation` pattern, since that endpoint can answer 404.
- Tests were added to `GridServiceTest` and `LocationServiceTest` covering, for each new method, the success path, the null-body path, the 404 translation where applicable, and the generic `RestClientException` wrapping.

No controller, DTO, OpenAPI spec, or Python SDK change was required — the three endpoints are already implemented by the controllers, already present in `docs/openapi/viron-api.json`, and already exposed by the Python SDK. This change closes the Java-side gap only.

## Test plan

- [x] `mvn -B test` — BUILD SUCCESS, 393 tests run, 0 failures, 0 errors (`GridServiceTest` 15, `LocationServiceTest` 54). Note that `./mvnw` was unavailable to this session's tooling, so the system `mvn` was used locally; the CI check on this PR head remains the authoritative anchor and invokes `./mvnw`.
- [x] The Python client was not modified, so `pytest` coverage is unaffected.
- [x] Each new method's URL template and verb were checked against `GridController` / `LocationController` and against `docs/openapi/viron-api.json`.

## Deferred work

Issue #130 (OpenAPI-first codegen / spec-drift guard) was not picked up this cycle: it calls for a build-tooling evaluation that changes `pom.xml` and the CI workflow, which is outside the scope of a polish-sized PR and warrants a human decision on the approach.

Closes #188

### DIFF
--- a/docs/PLANNING.md
+++ b/docs/PLANNING.md
@@ -32,7 +32,7 @@ Establish the core building blocks: project structure, models, DTOs, mapping, re
 Implement core service logic to support the REST API.
 
 11. **EnvironmentService** — Create, retrieve, update name, delete (cascade).  
-12. **GridService** — Lookups by ID, environment, or entity.  
+12. **GridService** — Lookups by ID, environment, or entity; update name.  
 13. **LocationService** — Manage locations, add/remove entities, lookups.  
 14. **EntityService** — Create, retrieve, delete entities and handle detach.
 
--- a/src/main/java/preponderous/viron/services/GridService.java
+++ b/src/main/java/preponderous/viron/services/GridService.java
@@ -3,13 +3,17 @@ package preponderous.viron.services;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.web.client.RestTemplateBuilder;
+import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Service;
 import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestTemplate;
 import preponderous.viron.config.AuthTokenInterceptor;
 import preponderous.viron.config.ServiceConfig;
+import preponderous.viron.dto.UpdateGridNameRequest;
+import preponderous.viron.exceptions.NotFoundException;
 import preponderous.viron.exceptions.ServiceException;
 import preponderous.viron.models.Grid;
 
@@ -82,6 +86,33 @@ public class GridService {
         } catch (Exception e) {
             log.error("Error getting grid for entity {}: {}", entityId, e.getMessage());
             throw new ServiceException("Failed to fetch grid for entity: " + entityId, e);
+        }
+    }
+
+    public boolean updateGridName(int id, String name) {
+        try {
+            RestTemplate restTemplate = restTemplateBuilder.build();
+            ResponseEntity<Void> response = restTemplate.exchange(
+                    baseUrl + "/{id}/name",
+                    HttpMethod.PATCH,
+                    new HttpEntity<>(new UpdateGridNameRequest(name)),
+                    Void.class,
+                    id
+            );
+            if (response.getStatusCode() == HttpStatus.NOT_FOUND) {
+                throw new NotFoundException("Grid not found with id: " + id);
+            }
+            if (!response.getStatusCode().is2xxSuccessful()) {
+                throw new ServiceException("Failed to update grid name");
+            }
+            return true;
+        } catch (HttpClientErrorException.NotFound e) {
+            throw new NotFoundException("Grid not found with id: " + id);
+        } catch (ServiceException e) {
+            throw e;
+        } catch (Exception e) {
+            log.error("Error updating name for grid {}: {}", id, e.getMessage());
+            throw new ServiceException("Failed to update grid name", e);
         }
     }
 }
--- a/src/main/java/preponderous/viron/services/LocationService.java
+++ b/src/main/java/preponderous/viron/services/LocationService.java
@@ -93,6 +93,22 @@ public class LocationService {
         }
     }
 
+    public List<Location> getUnoccupiedLocationsInGrid(int gridId) {
+        try {
+            ResponseEntity<Location[]> response = restTemplate.getForEntity(
+                    baseUrl + "/grid/{gridId}/unoccupied",
+                    Location[].class,
+                    gridId
+            );
+            return response.getStatusCode() == HttpStatus.OK && response.getBody() != null
+                    ? Arrays.asList(response.getBody())
+                    : Collections.emptyList();
+        } catch (Exception e) {
+            log.error("Error getting unoccupied locations in grid {}: {}", gridId, e.getMessage());
+            throw new ServiceException("Error getting unoccupied locations in grid: " + gridId, e);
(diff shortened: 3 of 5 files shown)


dmccoystephenson:
Self-review rubric (self-critique by the automation that opened this PR, not an independent review):

- **Scope: PASS** — Five files are touched: the two client services named in #188, their two test classes, and one line of `docs/PLANNING.md`. The docs line was added under the Phase 7 documentation check, since its sibling entry for `EnvironmentService` already lists "update name" and the `GridService` entry did not. No unrelated formatting, renames, or comment churn is present.
- **Tests-new: PASS** — All three new public methods are exercised: `updateGridName` by 5 tests, `getUnoccupiedLocationsInGrid` by 3, and `getNeighbors` by 4, covering success, null body, 404 translation where the endpoint can answer 404, and generic `RestClientException` wrapping.
- **Tests-fix: PASS (verified empirically, though this is a feature addition rather than a bug fix)** — Three deliberate mutations were applied (`updateGridName` verb changed `PATCH`→`PUT`; `getUnoccupiedLocationsInGrid` URL changed to the non-unoccupied route; `getNeighbors` URL changed to the entities route) and the suite went from 393/393 passing to 3 failures + 5 errors, all in the new tests. The mutations were reverted and the suite returned to 393/393. **One honest caveat, folded in here rather than inline:** `testUpdateGridName_NonSuccessStatus_ThrowsServiceException` and `testUpdateGridName_RestTemplateThrows_WrapsInServiceException` survived the verb mutation, because an unstubbed `exchange` returns `null` and the resulting `NullPointerException` is wrapped into a `ServiceException` carrying the same message the tests assert. Those two tests therefore do not independently pin the HTTP verb; `testUpdateGridName_Success_PatchesWithNameBody` does, via its `Mockito.verify` on `eq(HttpMethod.PATCH)`. The same shape exists in the pre-existing `EnvironmentServiceTest` tests they were modeled on.
- **Sibling structure: PASS** — No new files were created. The new methods sit in their existing classes, following the surrounding idiom: `GridService` rebuilds a `RestTemplate` per call from the builder, `LocationService` reuses the field-held instance, and each new test block follows the naming and section-comment style already used in its file.
- **Sibling renames: PASS** — No identifier was renamed. `updateGridName` was named to parallel the existing `updateEnvironmentName`, and `getUnoccupiedLocationsInGrid` / `getNeighbors` to parallel the Python SDK's `get_unoccupied_locations_in_grid` / `get_neighbors`.
- **Docs: PASS** — `docs/openapi/viron-api.json` already defines all three routes with the paths and verbs used here (verified by parsing the spec), so no contract change was needed. `docs/MVP.md`, `docs/REBUILD_PLAN.md`, `README.md`, and the Postman collection describe the HTTP surface, which is unchanged. `docs/PLANNING.md` was updated as noted above.
- **Issue resolution: PASS** — All three gaps enumerated in #188 are closed, and the issue's own scope statement (no controller, DTO, spec, or Python change) is respected.
- **CI: PASS** — `build` passed on head `c11bc37`.
- **DTO boundary: PASS** — No controller was touched. The client services deserialize into the `models` types (`Grid`, `Location`), which is the pre-existing convention for every sibling method in both classes.
- **Spec alignment: PASS** — `PATCH /api/v1/grids/{id}/name`, `GET /api/v1/locations/grid/{gridId}/unoccupied`, and `GET /api/v1/locations/{locationId}/neighbors` were each checked against both the controller method and the spec entry.
- **Java/Python parallelism: PASS** — This change is the parallelism fix: the Python SDK already exposed all three, and the Java client did not. No Python change is warranted.
- **Override correctness: PASS** — No `@Override` annotation is added or modified.

Two judgment calls are recorded rather than silently taken:

1. `src/main/java/preponderous/viron/services/GridService.java:92` — `updateGridName` maps a 404 to `NotFoundException`, whereas the other `GridService` methods map 404 to `Optional.empty()`. The cross-service parallel with `EnvironmentService.updateEnvironmentName` was preferred over intra-class consistency, since `Optional` carries no useful meaning for a write that returns no body. This matches what #188 proposed.
2. `src/main/java/preponderous/viron/services/LocationService.java:249` — `getNeighbors` returns an empty list on a non-`OK` status with a body, rather than treating it as an error. This follows `getLocationsInGrid` and the other list-returning siblings verbatim; diverging here would have been the less predictable choice.

One pre-existing observation outside this diff, noted for the record and not fixed here: `tickets.md` lists client method names that no longer match the code (`getGridsByEnvironmentId` vs. the implemented `getGridsInEnvironment`, `getLocationByEntityId` vs. `getLocationOfEntity`). That drift predates this PR and belongs in a separate documentation cycle.

PR #193, opened 2026-08-08
fix: enforce the single-placement invariant and unblock deleting a placed entity

### LINKED ISSUE #190: Adding an entity to a location returns 500 when the entity does not exist
## Summary

`PUT /api/v1/locations/{locationId}/entity/{entityId}` validates that the *location* exists but never validates that the *entity* exists.

`LocationController.addEntityToLocation` (`src/main/java/preponderous/viron/controllers/LocationController.java:69-77`) performs only a `locationRepository.findById(locationId)` guard before calling `locationRepository.addEntityToLocation(entityId, locationId)`.

## Observed behaviour

`LocationRepositoryImpl.addEntityToLocation` issues `INSERT INTO viron.entity_location (entity_id, location_id) VALUES (?, ?)`. The `entity_location` table declares `FOREIGN KEY (entity_id) REFERENCES viron.entity(entity_id)` (`db-scripts/setup/create_tables.sql:41`), so an unknown entity id violates the constraint. `DbInteractions.update` swallows the `SQLException`, logs it, and returns `false` (`src/main/java/preponderous/viron/database/DbInteractions.java:103-111`). The controller then throws `ServiceException`, which `GlobalExceptionHandler` maps to **500 Internal Server Error**.

A caller-supplied id that does not exist is a client error, so **404 Not Found** is the correct response — matching how every other resource in this repo reports a missing id.

## Proposed work

- An entity-existence guard should be added to `addEntityToLocation`, throwing `NotFoundException` when the entity id is unknown.
- The 404 response description in `docs/openapi/viron-api.json` for this path should be widened to cover a missing entity as well as a missing location.
- Controller tests covering the missing-entity case should be added to `LocationControllerTest`.

## Notes

The sibling `DELETE /api/v1/locations/{locationId}/entity/{entityId}` shares the same one-sided guard, but a delete of a non-existent association is harmless, so this issue is scoped to the `PUT`.

### LINKED ISSUE #191: An entity can be placed at two locations at once, breaking the single-placement invariant
## Summary

`PUT /api/v1/locations/{locationId}/entity/{entityId}` inserts an `entity_location` row unconditionally, with no check for an existing placement. The primary key of `viron.entity_location` is `(entity_id, location_id)` (`db-scripts/setup/create_tables.sql:40`), so nothing prevents the same entity from holding rows for two different locations simultaneously.

## Why this is a defect

The rest of the codebase assumes an entity occupies at most one location:

- `LocationRepositoryImpl.findByEntityId` uses `queryOne` and returns an `Optional<Location>` — with two placements, whichever row the database returns first wins and the other is silently invisible.
- `docs/MVP.md:49` describes `GET /api/v1/locations/entity/{entityId}` as "Retrieve **the** location of a specific entity" (singular).
- `LocationRepositoryImpl.moveEntityToLocation` issues `UPDATE viron.entity_location SET location_id = ? WHERE entity_id = ?`, which would collapse every duplicate placement onto one target.
- `LocationRepositoryImpl.removeEntityFromCurrentLocation` deletes every row for the entity, so a single `DELETE /api/v1/locations/entity/{entityId}` silently discards both placements.

A second symptom of the same missing guard: re-issuing the request for an entity **already at that same location** violates the composite primary key, which `DbInteractions.update` turns into `false` and the controller turns into `ServiceException` → **500**. `PUT` is expected to be idempotent, so a repeat of an already-satisfied placement should not fail.

## Proposed work

- A placement guard should be added to `LocationController.addEntityToLocation`:
  - if the entity is already at the requested location, the request should succeed as a no-op (idempotent `PUT`);
  - if the entity is already placed at a *different* location, `ConflictException` should be thrown (**409**), pointing callers at `PUT /{locationId}/entity/{entityId}/move` or `DELETE /entity/{entityId}`.
- A `409` response should be documented for this path in `docs/openapi/viron-api.json`.
- `LocationControllerTest` should cover the already-here, already-elsewhere, and unplaced cases.

## Notes

The move endpoint already rejects an occupied *target* with `ConflictException`, so `409` is the established status for a placement conflict in this repo.

### LINKED ISSUE #192: Deleting an entity that is placed at a location returns 500
## Summary

`DELETE /api/v1/entities/{id}` fails with **500 Internal Server Error** whenever the entity is currently placed at a location.

`EntityRepositoryImpl.deleteById` issues only `DELETE FROM viron.entity WHERE entity_id = ?` (`src/main/java/preponderous/viron/repositories/EntityRepositoryImpl.java:88-91`). The `entity_location` join table declares `FOREIGN KEY (entity_id) REFERENCES viron.entity(entity_id)` (`db-scripts/setup/create_tables.sql:41`), so any surviving placement row blocks the delete. `DbInteractions.update` swallows the resulting `SQLException` and returns `false`, `EntityController.deleteEntity` then throws `ServiceException`, and `GlobalExceptionHandler` maps that to 500.

An entity created through `POST /api/v1/entities` and then placed through `PUT /api/v1/locations/{locationId}/entity/{entityId}` therefore cannot be deleted at all through the API.

## Precedent in this repo

`EnvironmentController.deleteEnvironment` already handles exactly this ordering problem for the cascade it owns: `environmentRepository.deleteEntityLocation(entityId)` is called for every entity before `deleteEntity(entityId)` (`src/main/java/preponderous/viron/controllers/EnvironmentController.java:87-110`). The single-entity delete path was never given the same treatment.

## Proposed work

- `EntityRepositoryImpl.deleteById` should remove the entity's `entity_location` rows before deleting the entity row, returning the result of the entity delete.
- Repository tests should cover the placed-entity delete, and a controller test should confirm the endpoint returns 204 for a placed entity.
- The 500 documented in `docs/openapi/viron-api.json` for this path should be reviewed once the behaviour is corrected.

## Notes

No transaction wraps the two statements, since this repo uses raw JDBC via `DbInteractions` with no `@Transactional` usage anywhere. The failure window is narrow and strictly better than the current guaranteed failure, but it is called out here for visibility.

dmccoystephenson:
## Summary

Three defects around entity placement are fixed, all rooted in the same missing guards
between `viron.entity` and the `viron.entity_location` join table.

- **`PUT /api/v1/locations/{locationId}/entity/{entityId}` — entity existence (#190).**
  Only the location was validated before the insert, so an unknown entity id violated
  the `entity_location` foreign key, was swallowed by `DbInteractions.update`, and
  surfaced as a 500. An entity-existence guard is added, so an unknown entity id is now
  reported as **404**, matching how every other resource in this repo reports a missing id.

- **`PUT /api/v1/locations/{locationId}/entity/{entityId}` — single-placement invariant (#191).**
  The placement row was inserted unconditionally, so an entity could hold rows for two
  locations at once — a state that `findByEntityId`, `moveEntityToLocation` and
  `removeEntityFromCurrentLocation` all silently mishandle. A placement guard is added:
  an entity already at the requested location makes the request a **no-op success**
  (keeping `PUT` idempotent, where a repeat previously violated the composite primary key
  and became a 500), and an entity placed at a *different* location is rejected with
  **409**, the status this repo already uses for a placement conflict on the move endpoint.

- **`DELETE /api/v1/entities/{id}` — deleting a placed entity (#192).**
  `EntityRepositoryImpl.deleteById` issued only the entity delete, which any surviving
  placement row blocked via the foreign key, making a placed entity undeletable through
  the API. The entity's `entity_location` rows are now cleared first, mirroring the
  ordering `EnvironmentController.deleteEnvironment` already performs for the cascade it
  owns. The result of the entity delete alone is returned, since an unplaced entity has
  no placement row to clear.

Contract documentation and both client SDKs are updated in the same change, per this
repo's convention of keeping the spec in sync with the code:

- `docs/openapi/viron-api.json` — the 404 description is widened to cover a missing
  entity, a 409 response is documented, and the 200 description notes the no-op case.
  (No 500 was documented for `DELETE /api/v1/entities/{id}`, so nothing needed to be
  removed there.)
- `docs/MVP.md` — the endpoint line now describes the actual placement semantics.
- `services/LocationService.java` (Java client) and `services/locationService.py`
  (Python client) — both now translate a 409 rather than falling through to a generic
  failure.

## Test plan

- [x] `python3 -m pytest` — **107 passed**, covering the two new Python client cases.
- [ ] `./mvnw test -B` — **not runnable in this environment**: `java`, `mvn` and `./mvnw`
      are permission-denied by the sandbox this session runs in. The GitHub Actions
      `build` job on this branch is therefore the authoritative Java signal and is being
      relied upon in its place; merging is gated on it passing.
- New Java coverage added in this change:
  - `LocationControllerTest` — entity-not-found (404), already-at-target (no-op 200),
    already-elsewhere (409), unplaced-success, update-fails (500), and a
    `never()` assertion that the repository write is skipped on each rejected path.
  - `LocationServiceTest` — 409 status and `HttpClientErrorException.Conflict` both
    mapped to `ConflictException`.
  - `EntityRepositoryImplTest` — `InOrder` assertion that the placement delete precedes
    the entity delete, plus the unplaced-entity case confirming the outcome is taken from
    the entity delete alone.

A controller-level "204 for a placed entity" test was considered per #192 and deliberately
not added: `EntityController` reaches the database only through a mocked
`EntityRepository` in that test class, so such a test would be indistinguishable from the
existing `deleteEntity_ReturnsNoContent` case. The ordering fix is asserted at the
repository layer instead, which is where it lives.

## Known limitation

No transaction wraps the two statements in `deleteById`, as called out in #192 — this
repo uses raw JDBC through `DbInteractions` with no `@Transactional` usage anywhere. The
failure window is narrow and strictly better than the current guaranteed failure.

Closes #190
Closes #191
Closes #192

---

### DIFF
--- a/docs/MVP.md
+++ b/docs/MVP.md
@@ -50,7 +50,7 @@ Deliver a working, tested API that supports:
 - [x] `GET /api/v1/locations/{locationId}/entities` – Retrieve the IDs of the entities at a location.
 - [x] `GET /api/v1/locations/{locationId}/occupied` – Report whether a location holds any entities.
 - [x] `GET /api/v1/locations/{locationId}/neighbors` – Retrieve the locations adjacent to a location within its grid.
-- [x] `PUT /api/v1/locations/{locationId}/entity/{entityId}` – Add an entity to a location.
+- [x] `PUT /api/v1/locations/{locationId}/entity/{entityId}` – Place an unplaced entity at a location (no-op when the entity is already there, conflict when it is placed elsewhere).
 - [x] `PUT /api/v1/locations/{locationId}/entity/{entityId}/move` – Move a placed entity to an adjacent, unoccupied location in the same grid.
 - [x] `DELETE /api/v1/locations/{locationId}/entity/{entityId}` – Remove an entity from a specific location.
 - [x] `DELETE /api/v1/locations/entity/{entityId}` – Remove an entity from its current location.
--- a/docs/openapi/viron-api.json
+++ b/docs/openapi/viron-api.json
@@ -370,14 +370,21 @@
     "/api/v1/locations/{locationId}/entity/{entityId}": {
       "put": {
         "summary": "Add entity to location",
+        "description": "Places an unplaced entity at the location. An entity occupies at most one location, so the request is a no-op when the entity is already at that location and a conflict when it is placed elsewhere.",
         "parameters": [
           { "$ref": "#/components/parameters/LocationIdParam" },
           { "$ref": "#/components/parameters/EntityIdParam" }
         ],
         "responses": {
-          "200": { "description": "Entity added" },
+          "200": { "description": "Entity added, or already present at that location" },
           "404": {
-            "description": "Location not found with that id",
+            "description": "Location or entity not found with that id",
+            "content": {
+              "application/json": { "schema": { "$ref": "#/components/schemas/ErrorResponse" } }
+            }
+          },
+          "409": {
+            "description": "Entity is already placed at a different location",
             "content": {
               "application/json": { "schema": { "$ref": "#/components/schemas/ErrorResponse" } }
             }
--- a/src/main/java/preponderous/viron/controllers/LocationController.java
+++ b/src/main/java/preponderous/viron/controllers/LocationController.java
@@ -12,6 +12,7 @@ import preponderous.viron.exceptions.NotFoundException;
 import preponderous.viron.exceptions.ServiceException;
 import preponderous.viron.mappers.LocationMapper;
 import preponderous.viron.models.Location;
+import preponderous.viron.repositories.EntityRepository;
 import preponderous.viron.repositories.LocationRepository;
 
 import jakarta.validation.constraints.Min;
@@ -26,6 +27,7 @@ import java.util.stream.Collectors;
 @Validated
 public class LocationController {
     private final LocationRepository locationRepository;
+    private final EntityRepository entityRepository;
     private final LocationMapper locationMapper;
 
     @GetMapping
@@ -66,10 +68,26 @@ public class LocationController {
         return locationMapper.toDto(location);
     }
 
+    /**
+     * Places an unplaced entity at {@code locationId}. An entity occupies at most one location,
+     * so a request for an entity that is already placed elsewhere is a conflict; a request for an
+     * entity already at the target is a no-op, keeping the {@code PUT} idempotent.
+     */
     @PutMapping("/{locationId}/entity/{entityId}")
     public void addEntityToLocation(@PathVariable("entityId") @Min(1) int entityId, @PathVariable("locationId") @Min(1) int locationId) {
         if (locationRepository.findById(locationId).isEmpty()) {
             throw new NotFoundException("Location not found with id: " + locationId);
+        }
+        if (entityRepository.findById(entityId).isEmpty()) {
+            throw new NotFoundException("Entity not found with id: " + entityId);
+        }
+        Optional<Location> currentLocation = locationRepository.findByEntityId(entityId);
+        if (currentLocation.isPresent()) {
+            if (currentLocation.get().getLocationId() == locationId) {
+                return;
+            }
+            throw new ConflictException("Entity " + entityId + " is already placed at location "
+                    + currentLocation.get().getLocationId());
         }
         if (!locationRepository.addEntityToLocation(entityId, locationId)) {
             throw new ServiceException("Failed to add entity " + entityId + " to location " + locationId);
--- a/src/main/java/preponderous/viron/repositories/EntityRepositoryImpl.java
+++ b/src/main/java/preponderous/viron/repositories/EntityRepositoryImpl.java
@@ -84,10 +84,18 @@ public class EntityRepositoryImpl implements EntityRepository {
         return null;
     }
 
+    /**
+     * Deletes the entity and its placement, if any.
+     *
+     * <p>{@code viron.entity_location} carries a foreign key onto {@code viron.entity}, so the
+     * placement row has to be cleared first or the entity delete is rejected by the constraint.
+     * An entity that is not placed has no row to clear, which is why the first statement's result
+     * is not part of the outcome — only the entity delete itself is reported.
+     */
     @Override
     public boolean deleteById(int id) {
-        String query = "DELETE FROM viron.entity WHERE entity_id = ?";
-        return dbInteractions.update(query, id);
+        dbInteractions.update("DELETE FROM viron.entity_location WHERE entity_id = ?", id);
+        return dbInteractions.update("DELETE FROM viron.entity WHERE entity_id = ?", id);
     }
 
     @Override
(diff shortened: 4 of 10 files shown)


dmccoystephenson:
## Self-review

The full diff was read and checked against this repo's layering conventions, the three
issue specifications, and the surrounding call sites. No blocking defect was found. Two
non-blocking observations and four verification notes are recorded below.

Posted as a PR comment rather than a formal review object: neither `gh pr review` nor the
`pulls/<n>/reviews` API is available in this session, so each inline finding is folded in
below as a `path:line — finding` note.

### Observations (non-blocking)

`src/main/java/preponderous/viron/repositories/EntityRepositoryImpl.java:97` — the two
statements are not atomic, so a failure of the second (entity) delete leaves the entity
present but silently unplaced, having already lost its `entity_location` row. This is
accepted rather than fixed: no `@Transactional` usage exists anywhere in this repo, which
reaches the database exclusively through raw JDBC via `DbInteractions`, and introducing
transaction management for one method would be a larger architectural change than these
issues scope. The window is narrow and is strictly better than the guaranteed failure it
replaces, which is why #192 called it out for visibility rather than as a requirement.

`src/main/java/preponderous/viron/controllers/LocationController.java:84` — the placement
guard is a read-then-write, so two concurrent requests for the same unplaced entity can
both observe `Optional.empty()` and both proceed to the insert. The composite primary key
on `viron.entity_location` remains the real enforcement, so the invariant still holds; the
loser of that race receives a 500 rather than the 409 the guard would have produced. This
is a strict improvement over the previous behaviour (where every repeat was a 500) and
closing it fully would require the same transactional support noted above.

### Verified during review

`src/main/java/preponderous/viron/services/LocationService.java:146` — the new
`ConflictException` is thrown from inside the `try` block, which was checked against the
catch chain below it. Because `ConflictException` and `NotFoundException` both extend
`ServiceException`, the existing `catch (ServiceException e) { throw e; }` at line 156
rethrows them unchanged ahead of the general `catch (Exception e)`. Neither is rewrapped.

`src/main/java/preponderous/viron/controllers/DebugController.java:94` and `:142` — the
only other callers of `addEntityToLocation` were traced. Both create an entity immediately
before placing it, so the entity is always unplaced and the new 409 path is unreachable
from either. The guard is per-entity, not per-location, so the world generators placing two
entities on one location are also unaffected.

`src/test/java/preponderous/viron/repositories/EntityRepositoryImplTest.java:475-521` — the
three pre-existing `deleteById` tests were re-checked against the two-statement
implementation. Each stubs only the entity delete; the unstubbed placement delete returns
the mock default of `false`, and the outcome is taken from the second call, so all three
continue to assert what they did before.

`src/main/java/preponderous/viron/exceptions/GlobalExceptionHandler.java:23-28` — a handler
mapping `ConflictException` to a 409 `ErrorResponse` already exists, so no new handler was
required for the exception type this change begins throwing from a second endpoint.

### Verification status

`./mvnw test -B` could not be executed: `java`, `mvn` and `./mvnw` are permission-denied in
the sandbox this session runs in. The Java review above was therefore performed by reading
the sources rather than by running the suite, and the GitHub Actions `build` job is being
treated as the authoritative signal. `python3 -m pytest` was run locally and reports 107
passed.

---

This review comment was drafted during a Gardener session (https://github.com/Stephenson-Software/gardener).

Issue #194, opened 2026-08-08
Multi-statement writes are not atomic; a shared JDBC connection blocks adding transactions


dmccoystephenson:
## Summary

Several write paths in this service issue multiple dependent SQL statements that are not
atomic. A failure partway through leaves the database in a state that no single request can
produce deliberately, and no rollback occurs.

This was observed while reviewing PR #193 and is filed for visibility rather than as a
regression from that change — the gap predates it.

## Where this occurs

- `EntityRepositoryImpl.deleteById` (`src/main/java/preponderous/viron/repositories/EntityRepositoryImpl.java:97-98`)
  clears the entity's `viron.entity_location` row and then deletes the entity. If the second
  statement fails, the entity survives but has silently lost its placement.
- `EnvironmentController.deleteEnvironment` (`src/main/java/preponderous/viron/controllers/EnvironmentController.java:87-125`)
  performs a much longer cascade — association rows, then entities, then locations, then
  grids — throwing `ServiceException` from the middle of the loop on the first failure.
  Every deletion already performed is retained, leaving a partially deleted environment.
- `LocationController.addEntityToLocation` (`src/main/java/preponderous/viron/controllers/LocationController.java:78-94`)
  reads the location, the entity and the current placement before writing. Two concurrent
  requests for the same unplaced entity can both observe no placement and both attempt the
  insert. The composite primary key on `viron.entity_location` still enforces the invariant,
  so no corruption results, but the loser receives a 500 rather than the 409 the guard is
  meant to produce.

## Why this is not a one-line fix

`@Transactional` is not used anywhere in the codebase, and no transaction manager is
configured. Adding one is complicated by how `DbInteractions` is built
(`src/main/java/preponderous/viron/database/DbInteractions.java:26-33`): a single
`java.sql.Connection` is opened in the constructor and held as state on an application-scoped
`@Component`, and `setAutoCommit(false)` is never called.

That single shared connection means transaction boundaries cannot simply be switched on.
Calling `setAutoCommit(false)` on a connection shared by every concurrent request would
enclose unrelated requests in the same transaction, so one request's rollback would discard
another's committed-looking work. The shared mutable connection is also a thread-safety
concern in its own right, since a JDBC `Connection` is not required to be thread-safe.

## Proposed work

The prerequisite should be addressed first, then the boundaries:

1. `DbInteractions` should obtain connections from a pooled `DataSource` (Spring Boot
[...]

PR #198, opened 2026-08-09
refactor: pool database connections, and fix the blocked local-run configuration

### LINKED ISSUE #196: docker compose up fails: JWT_SECRET is required at startup but is not supplied by compose.yml or sample.env
## Summary

The documented local-run path (`docker-compose up --build`, per `README.md`) cannot start the
service, because a required environment variable is never supplied.

`src/main/resources/application.properties:25` declares the JWT secret with no default:

```
app.jwt.secret=${JWT_SECRET}
```

That placeholder is consumed by `SecurityConfig.jwtDecoder`
(`src/main/java/preponderous/viron/config/SecurityConfig.java:64-67`) via
`@Value("${app.jwt.secret}")`. When `JWT_SECRET` is unset, placeholder resolution fails and the
application context does not start.

## Where the variable is missing

- `compose.yml` — the `viron` service's `environment` block passes `DATABASE_DB_URL`,
  `DATABASE_DB_USERNAME`, `DATABASE_DB_PASSWORD`, `SERVICE_VIRON_HOST`, `SERVICE_VIRON_PORT` and
  `VIRON_DEBUG_ENABLED`, but no `JWT_SECRET`.
- `sample.env` — no `JWT_SECRET` entry exists, so a copy of it does not satisfy the requirement
  either.
- `README.md` — the Prerequisites and Running Locally sections do not mention that a JWT secret
  must be configured.

`JWT_ISSUER` and `JWT_ALGORITHM` do have defaults (`userauth` / `HS256`), so only the secret is
startup-blocking; the other two are worth documenting alongside it for completeness.

## Proposed work

- `JWT_SECRET` should be passed through in `compose.yml`'s `viron` service environment block.
- A placeholder `JWT_SECRET` (with a comment stating it must match the UserAuth service's secret
  and, for HS256, be at least 32 bytes) should be added to `sample.env`.
- The README's Getting Started section should state that the secret must be set before the
  service will start.

## Notes

The value placed in `sample.env` is a sample, not a credential; the existing sample database
password is treated the same way.

### LINKED ISSUE #197: README points at port 8080 while the service listens on 9999
## Summary

The README directs readers to port 8080, but the service listens on 9999.

`src/main/resources/application.properties:5` sets `server.port=9999`, and `compose.yml` publishes
`"9999:9999"` for the `viron` service. Two README statements disagree with that:

- `README.md:134` — "API will be available at: http://localhost:8080"
- `README.md:141` — "http://localhost:8080/swagger-ui.html"

Anyone following the Getting Started section reaches nothing on 8080.

## Proposed work

Both URLs should be corrected to port 9999, matching `server.port` and the published compose port.

## Notes

`sample.env` sets `SERVER_PORT=9999` as well, so 9999 is the intended value throughout and the
README is the only place carrying the stale port.

dmccoystephenson:
## Summary

Two independent changes are carried here: the connection-management prerequisite from #194,
and the local-run configuration gaps found during this cycle's triage.

### Pooled database connections (prerequisite for #194)

- `DbInteractions` no longer opens a single `java.sql.Connection` in its constructor and holds
  it as state on an application-scoped `@Component`. Every concurrent request shared that one
  connection, which JDBC does not require to be thread-safe.
- A HikariCP pool is now built from `DbConfig` by the new `config/DataSourceConfig`, and
  `query`/`queryOne`/`update` each borrow and release a connection per call.
- Connections are borrowed through `DataSourceUtils` rather than `DataSource.getConnection()`.
  Outside a transaction this is plain borrow-and-return; inside one, the transaction-bound
  connection is reused. This is what allows `@Transactional` boundaries to be placed around the
  multi-statement write paths described in #194 without revisiting this class.
- The pool is constructed with setters rather than `new HikariDataSource(HikariConfig)`, so
  pool start-up is deferred to first use and the context still starts when the database is
  unreachable — matching the previous behaviour.
- Actuator's database health contributor is disabled (`management.health.db.enabled=false`) so
  `/actuator/health` keeps its current meaning. Adding a `DataSource` bean would otherwise have
  made health report DOWN during a database outage, which is a readiness signal rather than the
  liveness signal the endpoint is used as today. The property is commented so it can be turned
  on deliberately.
- The unused `connect()` and `close()` methods are removed; the pool is closed by Spring. No
  caller existed outside the class.

This is step 1 of #194 only, which the issue itself suggests splitting out. Steps 2 and 3
(transaction boundaries, and turning the duplicate-placement race into a 409) remain open, so
#194 is not closed here.

### Local-run configuration (#196, #197)

- `JWT_SECRET` is now passed through in `compose.yml` and defined in `sample.env`, alongside
  `JWT_ISSUER` and `JWT_ALGORITHM`. Without it, `app.jwt.secret=${JWT_SECRET}` fails to resolve
  and the service does not start, so the README's documented `docker-compose up --build` path
  was broken.
- The README's Getting Started section now states that the secret is required.
- The README's two references to `http://localhost:8080` are corrected to `9999`, matching
  `server.port` and the port compose publishes.
- The README's Tech Stack entry records HikariCP pooling.

## Test plan

- [ ] `./mvnw compile -B` — run by CI; `java`/`mvn` are unavailable in the environment this
      change was authored in, so CI is the authoritative signal.
- [ ] `./mvnw test -B` — run by CI.
- New `DbInteractionsTest` cases cover connection release on every path (a pool of one with a
  short acquisition timeout, exercised across success and error paths) and concurrent callers
  each completing their own write.
- New `DataSourceConfigTest` covers construction from `DbConfig`, deferred pool start-up, and
  the context exposing exactly one pooled `DataSource` (so Spring Boot's auto-configured one
  has backed off).
- `pytest` was not run: the Python SDK is untouched, since no endpoint URL, request shape or
  response shape changed.

Closes #196.
Closes #197.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ The MVP implements the endpoints defined in `docs/openapi/viron-api.json` and do
 - Lombok
 - MapStruct (model ↔ DTO mapping)
 - Spring Security (OAuth2 resource server / JWT)
-- PostgreSQL (persistence layer)
+- PostgreSQL (persistence layer), pooled with HikariCP
 - Maven (build tool)
 - Docker + Docker Compose (deployment)
 - Swagger/OpenAPI (API documentation)
@@ -129,16 +129,22 @@ viron/
 ### Installation
 mvn clean install
 
+### Configuration
+Copy `sample.env` to `.env` and review the values before starting anything.  
+`JWT_SECRET` is required — it has no default, so the service will not start until it is set.  
+It must match the secret used by the UserAuth service that issues the tokens, and for `HS256`
+it must be at least 32 bytes.
+
 ### Running Locally
 docker-compose up --build  
-API will be available at: http://localhost:8080
+API will be available at: http://localhost:9999
 
 ---
 
 ## 📜 API Documentation
 
 Once running, you can view the interactive API docs:  
-http://localhost:8080/swagger-ui.html  
+http://localhost:9999/swagger-ui.html  
 or refer to the `docs/openapi/viron-api.json` file.
 
 ---
--- a/compose.yml
+++ b/compose.yml
@@ -27,6 +27,10 @@ services:
       SERVICE_VIRON_HOST: ${SERVICE_VIRON_HOST}
       SERVICE_VIRON_PORT: ${SERVICE_VIRON_PORT}
       VIRON_DEBUG_ENABLED: ${VIRON_DEBUG_ENABLED:-false}
+      # Required — the service does not start without it. Must match UserAuth's secret.
+      JWT_SECRET: ${JWT_SECRET}
+      JWT_ISSUER: ${JWT_ISSUER:-userauth}
+      JWT_ALGORITHM: ${JWT_ALGORITHM:-HS256}
     depends_on:
       - postgres
 
--- a/compose.yml
+++ b/compose.yml
@@ -28,7 +28,8 @@ services:
       SERVICE_VIRON_PORT: ${SERVICE_VIRON_PORT}
       VIRON_DEBUG_ENABLED: ${VIRON_DEBUG_ENABLED:-false}
       # Required — the service does not start without it. Must match UserAuth's secret.
-      JWT_SECRET: ${JWT_SECRET}
+      # The :? form makes compose refuse to start rather than pass an empty secret through.
+      JWT_SECRET: ${JWT_SECRET:?JWT_SECRET must be set; see sample.env}
       JWT_ISSUER: ${JWT_ISSUER:-userauth}
       JWT_ALGORITHM: ${JWT_ALGORITHM:-HS256}
     depends_on:
--- a/pom.xml
+++ b/pom.xml
@@ -23,6 +23,13 @@
             <artifactId>spring-boot-starter</artifactId>
         </dependency>
 
+        <!-- Supplies HikariCP (the connection pool behind DbInteractions) and spring-jdbc's
+             DataSourceUtils, which makes connection acquisition transaction-aware. -->
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-jdbc</artifactId>
+        </dependency>
+
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-test</artifactId>
--- a/sample.env
+++ b/sample.env
@@ -19,3 +19,10 @@ SERVICE_VIRON_PORT=9999
 # debug
 # Registers the /api/v1/debug data-seeding endpoints. Leave false outside local/dev.
 VIRON_DEBUG_ENABLED=false
+
+# jwt
+# Required: the service refuses to start without it. Must match the UserAuth service's
+# JWT_SECRET, and for HS256 must be at least 32 bytes. Replace the sample value below.
+JWT_SECRET=replace-me-with-the-userauth-secret-at-least-32-bytes
+JWT_ISSUER=userauth
+JWT_ALGORITHM=HS256
--- a/src/main/java/preponderous/viron/config/DataSourceConfig.java
+++ b/src/main/java/preponderous/viron/config/DataSourceConfig.java
@@ -1,0 +1,42 @@
+// Copyright (c) 2024 Preponderous Software
+// MIT License
+
+package preponderous.viron.config;
+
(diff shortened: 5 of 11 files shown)


dmccoystephenson:
## Self-review

The diff was re-read in full against the repository's conventions. One change is warranted and
is applied in a follow-up commit; the remaining notes record decisions that a reader of this
diff should not have to reconstruct. Inline findings are folded in below as `path:line` notes,
since this session cannot create a Review object with anchored comments.

### To change

`compose.yml:31 — JWT_SECRET: ${JWT_SECRET}`

When `JWT_SECRET` is unset in the environment or `.env`, Compose substitutes an empty string
rather than failing, so the service starts up and then dies inside `SecurityConfig.jwtDecoder`
when Nimbus rejects an empty HMAC key. Since the whole point of this hunk is to make the
documented local-run path fail comprehensibly, the required-variable form
(`${JWT_SECRET:?...}`) is a better fit: Compose then refuses to start and names the missing
variable. This is applied.

### Verified, no change needed

`src/main/java/preponderous/viron/database/DbInteractions.java:68,99,127 —
DataSourceUtils.getConnection/releaseConnection`

Each of the three entry points acquires before the `try` and releases in a `finally`, so the
connection is returned on the success path, on the `SQLException` path (including the early
`return new ArrayList<>()` inside `query`'s catch block, which the `finally` still follows), and
on any unchecked throw from a `RowMapper`. This is what the new
`everyPath_returnsItsConnectionToThePool` test pins down: a pool of one with a 1s acquisition
timeout turns any leaked connection into a failure on the next call.

`src/main/java/preponderous/viron/database/DbInteractions.java:38-41 — changed failure mode`

A database outage previously surfaced as a `NullPointerException` (the constructor logged the
connection failure and left the field null) or, mid-flight, as a swallowed `SQLException` that
became an empty result — and an empty result is indistinguishable from missing data, so a
controller would answer 404. Acquisition failure now raises
`CannotGetJdbcConnectionException`, which `GlobalExceptionHandler`'s catch-all turns into a 500.
The HTTP status for the constructor-failure case is unchanged; the mid-flight case is corrected
rather than preserved, and the javadoc says so.

`src/main/resources/application.properties:11-15 — management.health.db.enabled=false`

Introducing a `DataSource` bean makes Actuator register a database health contributor, which
would flip `/actuator/health` to DOWN (503) during a database outage. `/actuator/health` is the
only probe endpoint this service exposes and is public per `SecurityConfig`, so it is being
consumed as a liveness signal; letting it go DOWN would restart otherwise-healthy pods during a
database blip. Current behaviour is therefore preserved deliberately, with the property
commented so it can be enabled where readiness semantics are wanted instead. This is the one
consequence of the change that is worth a maintainer's opinion rather than a reviewer's.

`src/test/resources/application.properties:14-17 — duplicated property`

Not redundant: `src/test/resources/application.properties` shadows the main one entirely on the
test classpath rather than merging with it, so the setting has to appear in both files or
`SecurityConfigTest.actuatorHealthIsPublic` would start depending on a reachable database.

`src/main/java/preponderous/viron/config/DataSourceConfig.java:34-41 — setter-built pool`

`new HikariDataSource(HikariConfig)` starts the pool eagerly and fails fast, which would stop
the application context from starting whenever the database is unreachable — including in every
`@SpringBootTest` in this repository, none of which have a database. The setter form defers
start-up to the first `getConnection()`, preserving the previous behaviour;
`poolIsNotStartedUntilFirstConnectionIsRequested` guards that property so it is not lost to a
future refactor.

`DbInteractions.connect()` and `DbInteractions.close()` — removal

Both were public but unused: a search across `src/main` and `src/test` finds no caller, and the
pool is closed by the container through the bean's destroy method. The Python SDK is unaffected,
as it talks HTTP.

`src/test/java/preponderous/viron/config/DataSourceConfigTest.java:63 — single DataSource`

Worth keeping as an assertion rather than a comment: it is what proves Spring Boot's
`DataSourceAutoConfiguration` has backed off in favour of the explicit bean, which is otherwise
invisible and would silently produce a second, differently-configured pool if the explicit bean
were ever renamed out of the way.

### Scope

No unparameterised SQL, no new exception type, no DTO, and no endpoint change appear in this
diff, so the OpenAPI spec and the Python SDK are correctly untouched;
`OpenApiSpecDriftTest` continues to guard that. `@Transactional` is deliberately absent — this
is the connection-management prerequisite from #194 only, and that issue stays open for the
transaction boundaries and the 409-on-duplicate-placement work.

PR #202, opened 2026-08-12
fix: report the duplicate-placement loser as 409, not 500

### LINKED ISSUE #194: Multi-statement writes are not atomic; a shared JDBC connection blocks adding transactions

### LINKED ISSUE #200: Report the duplicate-placement loser in addEntityToLocation as 409, not 500
## Summary

`LocationController.addEntityToLocation` (`src/main/java/preponderous/viron/controllers/LocationController.java:76-95`) reads the location, the entity and the current placement before inserting. Two concurrent requests for the same unplaced entity can both observe no placement and both attempt the insert. The composite primary key on `viron.entity_location` still holds, so no corruption results, but the losing request receives a 500 where the guard immediately above it is meant to produce a 409.

This is step 3 of #194. Steps 1 (connection pooling, #198) and 2 (transaction boundaries, #199) are complete; this was split out of #199 rather than bundled with it.

## Why a transaction boundary alone does not close it

Wrapping the method in a transaction was considered while #199 was written and rejected as insufficient:

- Under the default READ COMMITTED isolation, two transactions can still each read no placement and then race on the insert. A boundary changes when the loser fails, not whether it fails.
- `SELECT ... FOR UPDATE` on the placement does not help either: there is no row to lock when the entity is unplaced, which is precisely the case that races.
- On Postgres a unique-key violation aborts the whole transaction, so re-reading the placement afterwards to decide "this was a conflict, not an error" fails inside the same boundary.

## Why it is not a small change

`DbInteractions.update` (`src/main/java/preponderous/viron/database/DbInteractions.java:126-137`) logs the `SQLException` and returns `false`. A duplicate-key violation is therefore indistinguishable from any other write failure by the time the controller sees it, which is what forces the 500. Reporting a 409 needs the constraint violation to be distinguishable, and each of the available routes has a cost worth weighing before one is picked:

1. Have `update` surface the `SQLException` (or a translated `DataAccessException`). This is the most faithful option but changes the error contract for every repository call in the service, so it needs its own review.
2. Make the insert conditional in SQL — `INSERT ... ON CONFLICT DO NOTHING` — so the losing request affects no rows and returns `false` without aborting the transaction. The controller can then re-read the placement and answer 409. This is narrower, but it moves the invariant into Postgres-specific SQL.
3. Translate at the boundary only, by catching the violation in the repository method that owns the insert.

## Acceptance

- Two concurrent placements of the same unplaced entity leave exactly one placed, and the loser receives 409 with a message naming the location the entity ended up at.
- A genuine write failure is still reported as 500, not misreported as a conflict.
- The behaviour is covered by a test that exercises the concurrent path rather than only the sequential guard.

---

dmccoystephenson:
## Summary

- `viron.entity_location` is now keyed on `entity_id` alone. The previous key,
  `(entity_id, location_id)`, does not express the invariant the rest of the service assumes —
  that an entity occupies at most one location — so two concurrent placements of the same
  unplaced entity at *different* locations both passed the read-then-write guard in
  `LocationController.addEntityToLocation` and both inserted, leaving the entity in two places
  at once. This is a correction to the premise recorded in #200 and #194, where the old key was
  described as holding in every case: it holds only when both requests name the same location.
  A regression test that reproduces the race fails against the old key with eight surviving
  placements.
- `DbInteractions.updateReportingDuplicateKey` was added, reporting a unique/primary-key
  violation (SQL state `23505`, as used by both Postgres and H2) as a
  `DuplicateKeyException` instead of the `false` that made a lost race indistinguishable from a
  genuine write failure and therefore forced a 500. `update` delegates to it and still flattens
  the violation to `false`, so the contract every pre-existing caller was written against is
  unchanged; only the placement insert opts in. This is option 3 of the three weighed in #200 —
  translation at the boundary that owns the insert — chosen over surfacing every `SQLException`
  (which would change the error contract service-wide) and over `INSERT ... ON CONFLICT DO
  NOTHING` (which would move the invariant into Postgres-only SQL and still infer the conflict
  from a re-read rather than observe it).
- The losing request is answered from the placement the winner committed: a 409 naming that
  location, and — where both requests named the same location — the same silent success the
  sequential path already returns, so the `PUT` stays idempotent whether the two arrive together
  or in sequence. This last point departs slightly from the acceptance criteria in #200, which
  ask for a 409 in every lost race; a 409 for a request whose desired outcome was reached would
  contradict the endpoint's documented no-op, so the guard's existing answer was preserved
  instead and is covered by its own test.
- A migration for databases created before this change is provided at
  `db-scripts/migrations/2026-08-12_entity_location_one_placement_per_entity.sql`. It is run by
  hand (the setup scripts execute only when Postgres initialises an empty volume) and deletes
  nothing: it aborts and reports the count if any entity is already placed twice, leaving the
  choice of which placement is real to an operator.
- The OpenAPI spec and the README were updated for the concurrent behaviour and the new
  migrations directory. The Python client SDK needed no change — `add_entity_to_location`
  already maps 409 to a conflict, and no endpoint's URL, request shape or response shape moved.

## Test plan

- [x] `mvn test -B` — 424 tests, 0 failures (up from 413; `./mvnw` is unavailable in this
      environment, so the system Maven 3.6.3 with Java 21 was used, matching what CI runs)
- [x] `ConcurrentPlacementTest` verified to fail without each half of the fix: with the old
      composite key, eight concurrent placements all survive; with the key in place but the
      violation still flattened to `false`, every loser is reported as a `ServiceException`
      (500). The interleaving is forced with a barrier held immediately after each request's
      placement read, since releasing threads together does not reproduce it — the winner
      commits before the others read.
- [x] `ConcurrentPlacementTest` run four times for flakiness; green each time
- [ ] pytest — not run; no Python SDK file was touched

Closes #200
Closes #194

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,8 @@ The MVP implements the endpoints defined in `docs/openapi/viron-api.json` and do
 - Docker + Docker Compose (deployment)
 - Swagger/OpenAPI (API documentation)
 - JaCoCo (test coverage)
-- Planned: Flyway (future migrations); schema currently comes from `db-scripts/setup/`
+- Planned: Flyway (future migrations); schema currently comes from `db-scripts/setup/`, with
+  changes to an already-created schema kept as hand-run scripts in `db-scripts/migrations/`
 
 ---
 
@@ -108,7 +109,7 @@ viron/
  │    └── services/          # Business logic  
  ├── src/main/python/        # Python client SDK  
  ├── src/test/java/...       # Unit and integration tests  
- ├── db-scripts/             # SQL schema setup scripts  
+ ├── db-scripts/             # SQL schema setup scripts, and migrations for existing databases  
  ├── docs/  
  │    ├── MVP.md             # Implementation checklist for MVP  
  │    └── openapi/  
--- a/db-scripts/migrations/2026-08-12_entity_location_one_placement_per_entity.sql
+++ b/db-scripts/migrations/2026-08-12_entity_location_one_placement_per_entity.sql
@@ -1,0 +1,47 @@
+-- Copyright (c) 2024 Preponderous Software
+-- MIT License
+
+-- Migration for issue #200: one placement per entity.
+--
+-- Databases created before this change key viron.entity_location on (entity_id, location_id),
+-- which permits the same entity to occupy two locations at once. Two concurrent placements of an
+-- unplaced entity at different locations both pass the controller's read-then-write guard and
+-- both insert, so the invariant the rest of the service assumes is not actually enforced.
+-- Keying on entity_id alone makes the database reject the second insert, which is what lets the
+-- losing request be reported as a 409 instead of a 500.
+--
+-- db-scripts/setup/create_tables.sql already creates new databases this way; this script brings
+-- an existing one into line. It is not applied automatically (the setup scripts run only when
+-- Postgres initialises an empty volume), so run it once against each existing database:
+--
+--     psql -U "$DATABASE_DB_USERNAME" -d "$DATABASE_DB_NAME" \
+--         -f db-scripts/migrations/2026-08-12_entity_location_one_placement_per_entity.sql
+--
+-- Nothing is deleted here. If any entity is already placed twice, the migration aborts and
+-- reports the count, leaving it to an operator to decide which placement is the real one.
+
+BEGIN;
+
+DO $$
+DECLARE
+    duplicate_count INT;
+BEGIN
+    SELECT count(*) INTO duplicate_count
+    FROM (
+        SELECT entity_id
+        FROM viron.entity_location
+        GROUP BY entity_id
+        HAVING count(*) > 1
+    ) duplicates;
+
+    IF duplicate_count > 0 THEN
+        RAISE EXCEPTION
+            'viron.entity_location places % entity/entities at more than one location; resolve them before applying this migration. To list them: SELECT entity_id, location_id FROM viron.entity_location WHERE entity_id IN (SELECT entity_id FROM viron.entity_location GROUP BY entity_id HAVING count(*) > 1) ORDER BY entity_id, location_id;',
+            duplicate_count;
+    END IF;
+END $$;
+
+ALTER TABLE viron.entity_location DROP CONSTRAINT entity_location_pkey;
+ALTER TABLE viron.entity_location ADD PRIMARY KEY (entity_id);
+
+COMMIT;
--- a/db-scripts/setup/create_tables.sql
+++ b/db-scripts/setup/create_tables.sql
@@ -34,10 +34,14 @@ CREATE TABLE IF NOT EXISTS viron.environment (
 );
 
 -- entity_location table (entity_id, location_id)
+-- An entity occupies at most one location, so entity_id alone is the key: a location may hold
+-- many entities, but a second placement of the same entity is rejected by the database rather
+-- than only by the read-then-write guard in LocationController.addEntityToLocation, which two
+-- concurrent requests can both pass.
 CREATE TABLE IF NOT EXISTS viron.entity_location (
     entity_id INT NOT NULL,
     location_id INT NOT NULL,
-    PRIMARY KEY (entity_id, location_id),
+    PRIMARY KEY (entity_id),
     FOREIGN KEY (entity_id) REFERENCES viron.entity(entity_id),
     FOREIGN KEY (location_id) REFERENCES viron.location(location_id)
 );
--- a/docs/openapi/viron-api.json
+++ b/docs/openapi/viron-api.json
@@ -370,7 +370,7 @@
     "/api/v1/locations/{locationId}/entity/{entityId}": {
       "put": {
         "summary": "Add entity to location",
-        "description": "Places an unplaced entity at the location. An entity occupies at most one location, so the request is a no-op when the entity is already at that location and a conflict when it is placed elsewhere.",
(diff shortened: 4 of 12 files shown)


dmccoystephenson:
## Self-review

The diff was read in full. No blocking problem was found; the notes below record what was
weighed, one accepted limitation, and one follow-up that is deliberately not part of this change.
(Posted as a comment rather than a formal review object, since the review API is not available to
this session.)

**Findings**

- `db-scripts/migrations/2026-08-12_entity_location_one_placement_per_entity.sql:44` — the drop
  names the constraint `entity_location_pkey`, which is the name Postgres derives for a table
  created by `db-scripts/setup/create_tables.sql`. A database whose key was created under a
  different name is not migrated by this script; the failure is loud and the whole script is
  wrapped in `BEGIN`/`COMMIT`, so nothing is half-applied. Left as is rather than made
  name-agnostic, because the only supported way this table gets created is the setup script.
- `src/main/java/preponderous/viron/database/DbInteractions.java:138` — a duplicate key reaching
  `update` is now logged as `Error executing update: Update violated a unique constraint: ...`,
  a slightly doubled prefix. Accepted: the message is still accurate and the alternative is
  duplicating the log call in the delegate.
- `src/main/java/preponderous/viron/database/DbInteractions.java:186` — detection is by SQL state
  rather than by exception type, because the Postgres driver reports `23505` on a `PSQLException`
  and not on the JDBC `SQLIntegrityConstraintViolationException` subclass. H2, which the tests
  run against, reports the same state, so the test coverage is meaningful for the production
  driver. The chain is walked with a self-reference guard so a malformed chain cannot hang the
  request thread.
- `src/main/java/preponderous/viron/controllers/LocationController.java:102` — the lost race is
  logged at INFO, not WARN or ERROR: it is an ordinary outcome of two clients competing, and the
  request is answered correctly.
- `src/main/java/preponderous/viron/controllers/LocationController.java:201` — out of scope and
  untouched here, but noted while reviewing: `moveEntityToLocation` has the same read-then-write
  shape, reading occupancy and then updating. Two entities moving into the same empty location
  concurrently can both pass that check, and unlike the placement invariant this one cannot be
  delegated to the database — a location is allowed to hold several entities (`addEntityToLocation`
  does not check occupancy at all), so no constraint expresses it. A separate issue is being filed
  rather than widening this change.

**Checked and found clean**

- No unparameterized SQL and no unclosed JDBC resource was introduced; the new throwing path
  releases its connection in the same `finally` as the others, which is asserted by a
  single-connection-pool test.
- No new exception type was added, so no `GlobalExceptionHandler` case is missing;
  `DuplicateKeyException` is caught where it is raised and never reaches the handler.
- No DTO or endpoint shape changed, so no validation or `@Schema` annotation is owed, and the
  Python client SDK contract is unaffected.
- Every new public method (`DbInteractions.updateReportingDuplicateKey`) is covered by tests, as
  is the changed behaviour of `update` on a duplicate key.
- The OpenAPI spec was updated in the same change, and `OpenApiSpecDriftTest` passes.

Issue #203, opened 2026-08-12
moveEntityToLocation's collision check is a read-then-write race the database cannot settle


dmccoystephenson:
## Summary

`LocationController.moveEntityToLocation` (`src/main/java/preponderous/viron/controllers/LocationController.java:196-206`) refuses a move into an occupied location:

```java
if (!locationRepository.getEntityIdsAtLocation(locationId).isEmpty()) {
    throw new ConflictException("Target location " + locationId + " is already occupied");
}
if (!locationRepository.moveEntityToLocation(entityId, locationId)) {
```

The read and the write are separate statements with no boundary between them, so two entities moved into the same empty location at the same time both see it empty and both commit. Both end up there, and the collision the check exists to prevent has happened anyway.

This was noticed while #200 was being fixed and is filed separately rather than folded in, because the two cases are not the same shape.

## Why the fix used for #200 does not apply

The placement race in #200 was settled by giving `viron.entity_location` a primary key on `entity_id`, so the database itself rejects a second placement of one entity. There is no equivalent constraint available here: a location is allowed to hold several entities, and `addEntityToLocation` places an entity without consulting occupancy at all. "At most one entity per location" is therefore not an invariant of the data — it is a rule that only this one endpoint applies — and a unique key on `location_id` would contradict the rest of the service.

## Options worth weighing

1. Decide whether one-entity-per-location is meant to be a real invariant. If it is, `addEntityToLocation` is inconsistent with it and the schema can enforce it; if it is not, the check in `moveEntityToLocation` is a game rule that a spatial-state service may not be the right place for.
2. Serialize the move against the target location, for example by locking the target's rows (`SELECT ... FROM viron.entity_location WHERE location_id = ? FOR UPDATE`) inside a `@Transactional` boundary. This works where #200's version did not, because here there may well be rows to lock — though it still does not serialise against a concurrent move into a location that is genuinely empty.
3. Make the update conditional in SQL, so the write itself asserts the emptiness it depends on rather than trusting an earlier read.

## Impact

Two entities can share a location after concurrent moves. No constraint is violated and nothing needs manual repair, since the schema permits it; what is lost is the guarantee the endpoint's 409 appears to offer.

## Acceptance

- Two concurrent moves into the same empty location leave the location in a state the chosen answer to option 1 allows, and the losing request is told so.
- The behaviour is covered by a test that exercises the concurrent path, in the manner of `ConcurrentPlacementTest`.

[...]

PR #205, opened 2026-08-12
Make the Python client test suite runnable and document how to run it

### LINKED ISSUE #201: pytest fails to collect the Python client tests: pytest.ini's pythonpath does not match the tests' import style
## Summary

Running `pytest` from the repository root fails during collection for all eight Python test modules:

```
src/test/python/preponderous/viron/models/test_entity.py:6: in <module>
    from src.main.python.preponderous.viron.models.entity import Entity
E   ModuleNotFoundError: No module named 'src'
```

## Cause

The two settings disagree about how the client is imported.

`pytest.ini` puts the package directories themselves on the path:

```
[pytest]
pythonpath = ./src/main/python
             ./src/test/python
```

which supports `from preponderous.viron.models.entity import Entity`. Every test module instead imports through the full path from the repository root, for example `src/test/python/preponderous/viron/models/test_entity.py:6`:

```python
from src.main.python.preponderous.viron.models.entity import Entity
```

which needs the repository root on the path — and the root is never added. There are no `__init__.py` files under `src/test/python`, so pytest's default prepend import mode inserts each test module's own directory rather than the root, and `src` stays unresolvable.

## Impact

Developer experience and the loop's ability to verify client changes, not correctness of shipped code. The tests themselves are fine: all 107 pass once the path is right (`pytest --import-mode=importlib` is one way through). CI does not cover the Python client — `.github/workflows/ci.yml` runs only `./mvnw compile -B` and `./mvnw test -B` — so nothing catches this, and the documented command simply does not work for a newcomer.

## Suggested fix

Pick one import style and make the configuration agree with it. Adding the root:

```
pythonpath = .
```

makes the existing `src.main.python....` imports resolve with no change to any test. Rewriting the imports to `preponderous.viron....` and keeping the current `pythonpath` is the tidier option but touches all eight modules, and it is worth deciding deliberately since the package layout is mirrored from the Java side.

Worth resolving alongside a decision about whether CI should run the Python suite at all; today a green CI says nothing about the client.

---

### LINKED ISSUE #204: environmentService's get_all_environments uses a builtin generic annotation the rest of the client avoids, breaking Python 3.8
## Summary

`src/main/python/preponderous/viron/services/environmentService.py:21` annotates its return type with a PEP 585 builtin generic:

```python
def get_all_environments(self) -> list[Environment]:
```

Every other service in the client uses `typing.List` for the same shape — `entityService.py:21`, `gridService.py:24`, `locationService.py:19`. `environmentService.py` is also the only service whose `typing` import omits `List` (`from typing import Optional`).

## Impact

Builtin generics in annotations are evaluated at class-definition time and are only subscriptable from Python 3.9 onward, so importing the module under Python 3.8 raises at import:

```
src/main/python/preponderous/viron/services/environmentService.py:21: in EnvironmentService
    def get_all_environments(self) -> list[Environment]:
E   TypeError: 'type' object is not subscriptable
```

The whole `EnvironmentService` class — not merely the one method — becomes unimportable, and `src/test/python/preponderous/viron/services/test_environmentService.py` cannot be collected. No minimum Python version is declared anywhere in the repository (`requirements.txt` pins only `pytest` and `requests`; there is no `setup.py` or `pyproject.toml`), so the raised floor is undocumented as well as unintended.

## Suggested fix

The annotation can be brought in line with its siblings:

```python
from typing import List, Optional
...
def get_all_environments(self) -> List[Environment]:
```

This is behaviour-preserving on Python 3.9+ and restores importability on 3.8.

---

dmccoystephenson:
## Summary

- `pytest.ini`'s `pythonpath` is pointed at the repository root. The previous value listed the package directories themselves, which supports `preponderous.viron...` imports, whereas the client source and all eight test modules import through `src.main.python.preponderous.viron...` — so every test module failed collection with `ModuleNotFoundError: No module named 'src'`. Of the two ways out described in #201, the root was chosen deliberately: the alternative (rewriting the imports to `preponderous.viron...`) would have to change the client's own internal imports too, and that is the path external consumers import by, so it is an API decision rather than a test-configuration one. `testpaths` is also pinned so a bare `pytest` run picks up only the client suite.
- `environmentService.get_all_environments` is annotated with `typing.List` instead of the builtin generic `list[Environment]`, matching all three sibling services. Builtin generics in annotations are evaluated when the class body runs and are not subscriptable before Python 3.9, so `EnvironmentService` could not be imported at all on 3.8 and `test_environmentService.py` could not be collected.
- The README's Testing section now documents the client suite, its Python and dependency requirements, why the root is on the path, and that CI covers the Java build only. The project-structure listing gains `src/test/python/`.

## Test plan

- [x] `mvn -B test` — 424 tests, 0 failures, 0 errors, BUILD SUCCESS
- [x] `pytest` (pytest 9.0.3) — 107 passed
- [x] `python3 -m pytest` (Python 3.8.10, pytest 7.1.3) — 107 passed, confirming the annotation fix on the lowest interpreter available here
- [x] Regression evidence for #201: with `pytest.ini` stashed, `pytest` reports `Interrupted: 8 errors during collection`; restored, 107 pass
- [x] Regression evidence for #204: with `environmentService.py` stashed, `python3 -m pytest` reports `TypeError: 'type' object is not subscriptable` collecting `test_environmentService.py`; restored, 107 pass. Note this failure is only observable on Python 3.8 — on 3.9+ the reverted annotation is valid, so the suite alone does not guard it.

## Notes on the rest of the backlog

Deferred this cycle, with reasons recorded here per the loop's auditability rule:

- **#203** (`moveEntityToLocation`'s read-then-write race) — the issue's own acceptance criteria are gated on a design decision that is not the loop's to make: whether one-entity-per-location is meant to be a real invariant of the data or a rule of that one endpoint. `addEntityToLocation` currently ignores occupancy entirely, so either answer implies changes beyond the reported endpoint. A conditional-write fix (option 3) is implementable without settling that question and is the recommended next step once the question is answered.
- **#130** (OpenAPI-first codegen / spec-drift guard) — build and CI infrastructure touching `pom.xml` and `.github/workflows/`, both on the do-not-auto-merge list, and larger than a polish-sized PR.

CI runs `./mvnw compile` and `./mvnw test` only, so a green run does not exercise any file changed here except by proving the Java build is unaffected; the Python verification above was performed locally.

Closes #201
Closes #204

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,7 @@ from tests as it is from application code (`src.main.python.preponderous.viron..
 `pythonpath` setting it uses arrived in pytest 7.0, so the pytest 6.2.4 pinned by
 `requirements.txt` ignores it and collection fails; install a newer pytest until that pin is
 raised.
+
 Note that CI runs the Java build and tests only — Python client changes are not covered there
 and must be checked locally.
 
--- a/README.md
+++ b/README.md
@@ -109,6 +109,7 @@ viron/
  │    └── services/          # Business logic  
  ├── src/main/python/        # Python client SDK  
  ├── src/test/java/...       # Unit and integration tests  
+ ├── src/test/python/...     # Python client SDK tests (pytest)  
  ├── db-scripts/             # SQL schema setup scripts, and migrations for existing databases  
  ├── docs/  
  │    ├── MVP.md             # Implementation checklist for MVP  
@@ -152,8 +153,16 @@ or refer to the `docs/openapi/viron-api.json` file.
 
 ## 🧪 Testing
 
-Run all unit and integration tests:  
+Run all Java unit and integration tests:  
 mvn test
+
+Run the Python client tests (requires Python 3.8+ and `pip install -r requirements.txt`):  
+pytest
+
+`pytest.ini` puts the repository root on the path, so the client is imported the same way
+from tests as it is from application code (`src.main.python.preponderous.viron...`).
+Note that CI runs the Java build and tests only — Python client changes are not covered there
+and must be checked locally.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -156,11 +156,14 @@ or refer to the `docs/openapi/viron-api.json` file.
 Run all Java unit and integration tests:  
 mvn test
 
-Run the Python client tests (requires Python 3.8+ and `pip install -r requirements.txt`):  
+Run the Python client tests (requires Python 3.8+, `requests`, and pytest 7 or newer):  
 pytest
 
 `pytest.ini` puts the repository root on the path, so the client is imported the same way
-from tests as it is from application code (`src.main.python.preponderous.viron...`).
+from tests as it is from application code (`src.main.python.preponderous.viron...`). The
+`pythonpath` setting it uses arrived in pytest 7.0, so the pytest 6.2.4 pinned by
+`requirements.txt` ignores it and collection fails; install a newer pytest until that pin is
+raised.
 Note that CI runs the Java build and tests only — Python client changes are not covered there
 and must be checked locally.
 
--- a/pytest.ini
+++ b/pytest.ini
@@ -3,4 +3,3 @@
 # root (src.main.python.preponderous.viron...), so the root is what belongs on
 # the path.
 pythonpath = .
-testpaths = src/test/python
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,3 +1,6 @@
 [pytest]
-pythonpath = ./src/main/python
-             ./src/test/python
+# The client and its tests import through the full path from the repository
+# root (src.main.python.preponderous.viron...), so the root is what belongs on
+# the path.
+pythonpath = .
+testpaths = src/test/python
--- a/src/main/python/preponderous/viron/services/environmentService.py
+++ b/src/main/python/preponderous/viron/services/environmentService.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2024 Preponderous Software
 # MIT License
 
-from typing import Optional
+from typing import List, Optional
 import requests
 from src.main.python.preponderous.viron.models.environment import Environment
 
@@ -18,7 +18,7 @@ class EnvironmentService:
     def get_auth_headers(self) -> dict:
         return {"Authorization": f"Bearer {self.auth_token}"} if self.auth_token else {}
 
-    def get_all_environments(self) -> list[Environment]:
+    def get_all_environments(self) -> List[Environment]:
         response = requests.get(f"{self.get_base_url()}", headers=self.get_auth_headers())
         response.raise_for_status()
         return [Environment(**env) for env in response.json()]


dmccoystephenson:
Self-review rubric, scored against the diff at `29853d6` and the command output named in each line. Items are scored FAIL unless direct evidence of PASS was produced.

- **CI (external anchor):** PASS — `gh pr checks 205 --watch` reports `build pass` on the current head. Scope caveat: the `CI` workflow runs `./mvnw compile -B` and `./mvnw test -B` only, so it exercises none of the three files changed here; what it proves is that the Java build is unaffected. The Python verification below was run locally and is the real anchor for this PR. That structural gap is now filed as #207.
- **Scope:** FAIL, then fixed — the first push also added `testpaths = src/test/python` to `pytest.ini`, which was not needed for either issue. `pytest -q -o "testpaths="` collects the same 107 tests without it, so the line changed nothing measurable and was removed in `b3b8fe3`. Re-scored PASS: the remaining diff is one path setting, one annotation plus its import, and the README.
- **Tests-new:** PASS by inapplicability — no public method or function is added. `get_all_environments`, whose annotation changed, is already exercised by `test_environmentService.py:20`, `:181`, and `:199`.
- **Tests-fix (empirical):** PASS — `git stash push -- pytest.ini` makes `pytest` report `Interrupted: 8 errors during collection`, and restoring it returns 107 passed. `git stash push -- src/main/python/.../environmentService.py` makes `python3 -m pytest` (Python 3.8.10, pytest 7.1.3) report `TypeError: 'type' object is not subscriptable` while collecting `test_environmentService.py`, and restoring it returns 107 passed. Stated plainly: the second failure is observable **only** below Python 3.9 — on 3.9+ the reverted annotation is legal, so the suite by itself does not guard #204 unless CI runs a 3.8 interpreter.
- **Sibling structure:** PASS by inapplicability — no new file is created.
- **Sibling renames:** PASS — `grep -rn "list\[" src/main/python/` now returns nothing, so all four services annotate list returns with `typing.List` consistently.
- **Docs:** PASS — the Phase 7 sweep was run over every source of truth. `docs/openapi/viron-api.json`, `docs/MVP.md`, `docs/PLANNING.md`, `docs/REBUILD_PLAN.md`, `tickets.md`, and `postman/` contain no reference to Python, pytest, or the client's import path (`grep -rn -i python` across `docs/`, `tickets.md`, and `.github/` matches nothing), and no endpoint, DTO, or schema is touched by this diff. `README.md` is the only affected doc and is updated.
- **Issue resolution:** PASS with one deliberate exclusion — #201's core claim (the `pythonpath`/import-style mismatch) is fixed, and #204's named annotation is changed. #201 additionally muses that this is "worth resolving alongside a decision about whether CI should run the Python suite at all"; that aside is not resolved here and is filed as #207 rather than left implied, so closing #201 does not quietly absorb it.
- **DTO boundary / Spec alignment / Java-Python parallelism / Override correctness:** PASS by inapplicability — no Java file, controller, endpoint, or DTO is touched. The annotation change has no Java counterpart, since the Java side already returns `List<EnvironmentDto>`.

## Findings folded in from outside the rubric

- `requirements.txt` pins `pytest==6.2.4`, which predates the `pythonpath` ini setting (introduced in pytest 7.0). A developer installing from that file therefore still hits the original collection failure, because the key is silently ignored. This was **not** verified by execution — a virtualenv could not be created in this environment (`ensurepip is not available`), so the claim rests on pytest's changelog rather than an observed run, and it is asserted in the README on that basis. The pin itself is left alone because `requirements.txt` is a dependency surface this repository's automation holds back from autonomous change; it is filed as #206 and the README carries an explicit warning in the meantime.
- The two verified interpreter/pytest combinations differ, which is worth stating so later readers are not misled: `pytest` on this machine is 9.0.3 running on 3.9+, while `python3 -m pytest` is 7.1.3 on 3.8.10. Both pass all 107 tests with this branch applied.

One intrinsic-critique pass; no further self-scoring will follow absent a reviewer comment or a CI failure.

---

PR #208, opened 2026-08-14
build: raise the pytest floor and cover the Python client in CI

### LINKED ISSUE #206: requirements.txt pins a pytest older than the pythonpath setting the test suite depends on
## Summary

`requirements.txt` pins:

```
pytest==6.2.4
requests==2.25.1
```

`pytest.ini` configures the import path with the `pythonpath` setting, which was introduced in pytest 7.0. Under the pinned pytest 6.2.4 that key is an unrecognised ini option and is ignored, so the repository root never reaches `sys.path` and collection of the client tests fails the same way it did before #201 was fixed — a developer following `pip install -r requirements.txt` lands back on the original error.

## Scope

The pin predates the `pythonpath` setting itself, so this is not a regression introduced by #201's fix; that fix simply makes the stale pin the remaining obstacle. The README now warns about it as a stopgap, which is documentation standing in for a dependency change.

## Suggested fix

Raise the pin to a pytest that provides `pythonpath` (7.0 or newer; the suite has been observed passing on 7.1.3 under Python 3.8.10 and on 9.0.3). Whether `requests` should move off 2.25.1 at the same time, and whether the two pins should be exact or floors, is a dependency-policy decision worth making deliberately — hence an issue rather than a drive-by bump, since `requirements.txt` is also the client's declared dependency surface.

Worth considering alongside the absence of any declared minimum Python version (no `setup.py` or `pyproject.toml`), noted in #204.

---

### LINKED ISSUE #207: CI does not run the Python client test suite, so client regressions are caught only by hand
## Summary

`.github/workflows/ci.yml` runs two steps, `./mvnw compile -B` and `./mvnw test -B`. Nothing in CI collects or runs the 107 tests under `src/test/python`, so a green check on a pull request says nothing about the Python client under `src/main/python`.

Two defects found this cycle illustrate the cost. The `pytest.ini` path mismatch fixed by #201 left every client test module uncollectable, and the builtin-generic annotation fixed by #204 made `EnvironmentService` unimportable on Python 3.8; both survived in `main` because CI never exercised that surface. This aside was raised in #201 and is filed separately so the fix there is not held up by it.

## Suggested work

A step that installs the client's dependencies and runs `pytest` would close the gap. Two decisions are worth making deliberately rather than assumed:

1. **Which interpreters.** No minimum Python version is declared anywhere (see #204), so the matrix is currently undefined. Pinning one version is simpler; a small matrix would have caught #204 directly.
2. **Which pytest.** `requirements.txt` pins a pytest that predates the `pythonpath` setting `pytest.ini` relies on (see #206), so a CI step installing from that file as it stands would fail collection.

Resolving #206 first would leave this a straightforward addition. Note that `.github/workflows/` is a path this repository's automation holds back from autonomous merge, so the change wants human review regardless.

---

dmccoystephenson:
## Summary

- `requirements.txt` asked for pytest 6.2.4, a release that predates the `pythonpath` ini setting `pytest.ini` depends on. Anyone installing from that file was returned to the collection failure fixed in #201, which the README had been papering over with a warning. The floor is raised to pytest 7.0.
- Both entries are expressed as floors rather than exact pins, because this file doubles as the declared dependency surface of the client under `src/main/python`, where exact pins fight with consuming applications. `requests` moves off 2.25.1 (December 2020) to 2.32.0 at the same time.
- A `python-client` job is added to `.github/workflows/ci.yml`. It installs `requirements.txt` and runs `pytest`, so the 107 tests under `src/test/python` are exercised on every push and pull request instead of by hand.
- The matrix covers Python 3.8 — the oldest interpreter the README claims — and 3.12. That shape is what would have caught the builtin generic annotation fixed in #204, which was invisible to CI at the time.
- The README's testing section is updated: the stopgap warning about the stale pin is removed, and the note that CI covers the Java build alone no longer holds.

## Test plan

- [x] `pytest` — 107 passed locally (Python 3.10.12, pytest 9.1.1)
- [x] No Java sources were touched by this change, so `./mvnw test -B` is left to the `build` job on this branch for its authoritative signal
- [x] The client suite mocks `requests.get`/`post`/`patch`/`delete` throughout, so the version bump is not exercised by test behaviour
- [x] The client and its tests were checked for PEP 585 builtin generics and PEP 604 unions, of which none remain, so the 3.8 leg is expected to be green

Closes #206
Closes #207

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,3 +25,30 @@ jobs:
 
       - name: Run tests
         run: ./mvnw test -B
+
+  python-client:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # The oldest interpreter the client claims to support and a current
+        # one, so syntax or typing that only newer releases accept is caught.
+        python-version: [ '3.8', '3.12' ]
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: pip
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install -r requirements.txt
+
+      - name: Run tests
+        run: python -m pytest
--- a/README.md
+++ b/README.md
@@ -156,17 +156,19 @@ or refer to the `docs/openapi/viron-api.json` file.
 Run all Java unit and integration tests:  
 mvn test
 
-Run the Python client tests (requires Python 3.8+, `requests`, and pytest 7 or newer):  
+Install the Python client's dependencies (requires Python 3.8+):  
+pip install -r requirements.txt
+
+Run the Python client tests:  
 pytest
 
 `pytest.ini` puts the repository root on the path, so the client is imported the same way
 from tests as it is from application code (`src.main.python.preponderous.viron...`). The
-`pythonpath` setting it uses arrived in pytest 7.0, so the pytest 6.2.4 pinned by
-`requirements.txt` ignores it and collection fails; install a newer pytest until that pin is
-raised.
+`pythonpath` setting it uses arrived in pytest 7.0, which is why `requirements.txt` asks for
+that release or newer.
 
-Note that CI runs the Java build and tests only — Python client changes are not covered there
-and must be checked locally.
+CI runs both suites: the Java build and tests, and the Python client tests on Python 3.8 and
+3.12.
 
 ---
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,10 @@
-pytest==6.2.4
-requests==2.25.1
+# Dependencies of the Python client under src/main/python and its pytest suite.
+# Floors rather than exact pins: this file doubles as the client's declared
+# dependency surface, and exact pins there fight with the applications that
+# consume it.
+#
+# pytest 7.0 is the release that introduced the `pythonpath` ini setting that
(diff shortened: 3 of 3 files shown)


dmccoystephenson:
## Self-review

The diff was read in full and reviewed against this repository's conventions. No blocking issues were found; three points are recorded below as notes rather than requested changes, since each reflects a deliberate choice made while resolving #206 and #207.

`requirements.txt:9-10` — floors (`>=`) replace exact pins. The reproducibility that exact pins buy is given up here: a future pytest or requests release can turn CI red without any change landing in this repository. That trade was taken because the same file is the client's declared dependency surface, where exact pins conflict with consuming applications, and because no lock file exists to hold a resolved set. Should the reproducibility matter more later, the conventional answer is a separate `requirements-dev.txt` or a lock file rather than a return to exact pins.

`.github/workflows/ci.yml:38` — Python 3.8 reached end of life in October 2024 and is kept in the matrix only because the README declares it as the floor. Nothing in the repository enforces that floor (no `setup.py` or `pyproject.toml`, as noted in #204), so the matrix is currently the only place the claim is checked. If the floor is raised, this entry is what should move with it.

`.github/workflows/ci.yml:52` — `python -m pytest` is used rather than a bare `pytest`, matching how the README's `pytest.ini` explanation depends on the invoking directory being the repository root. Both jobs confirmed collection of all 107 tests rather than an empty run, which is the failure mode a green pytest step can otherwise hide.

Verification on this branch: `build` passed, `python-client (3.8)` collected 107 items and passed on pytest 8.3.5 with requests 2.32.4, and `python-client (3.12)` collected 107 items and passed. The 3.8 leg confirms both that the interpreter is still installable on `ubuntu-latest` and that the raised floors resolve there.

PR #211, opened 2026-08-19
fix: settle the move collision check against concurrent moves

### LINKED ISSUE #203: moveEntityToLocation's collision check is a read-then-write race the database cannot settle

### LINKED ISSUE #210: Removing an entity that is not at the given location answers 500, not 404
## Summary

`DELETE /api/v1/locations/{locationId}/entity/{entityId}` answers `500` when the entity named is not actually at the location named. `LocationController.removeEntityFromLocation` (`src/main/java/preponderous/viron/controllers/LocationController.java:126-135`) guards only the location's existence:

```java
if (locationRepository.findById(locationId).isEmpty()) {
    throw new NotFoundException("Location not found with id: " + locationId);
}
if (!locationRepository.removeEntityFromLocation(entityId, locationId)) {
    throw new ServiceException("Failed to remove entity " + entityId + " from location " + locationId);
}
```

The repository statement is `DELETE FROM viron.entity_location WHERE entity_id = ? AND location_id = ?` (`src/main/java/preponderous/viron/repositories/LocationRepositoryImpl.java:76-79`), and `DbInteractions.update` returns `false` when no row was affected. A request naming an entity that does not exist, or one that exists but is placed somewhere else, therefore affects no rows and is reported as `ServiceException`, which `GlobalExceptionHandler` maps to `500 Internal Server Error` (`src/main/java/preponderous/viron/exceptions/GlobalExceptionHandler.java:58-62`).

A caller mistake is being answered as a server fault.

## What the contract says

`docs/openapi/viron-api.json` documents only `204` and `404 Location not found with that id` for this operation. The `500` this path can produce is undocumented, and the case that produces it — the entity is not at that location — has no documented answer at all.

The Python client already expects the missing answer to be a `404`: `remove_entity_from_location` (`src/main/python/preponderous/viron/services/locationService.py:68-72`) raises `"Location or entity not found"` on `404` and otherwise calls `raise_for_status()`, so today it surfaces this case as a raw HTTP 500.

## Precedent within the service

The sibling endpoint `DELETE /api/v1/locations/entity/{entityId}` already answers `404` for the equivalent situation: `removeEntityFromCurrentLocation` (`src/main/java/preponderous/viron/controllers/LocationController.java:137-146`) checks `findByEntityId(entityId).isEmpty()` and throws `NotFoundException` before attempting the delete. The two endpoints disagree about the same fact.

## Not covered by tests

`LocationControllerTest` covers success, location-not-found, and repository-throws for this endpoint (`src/test/java/preponderous/viron/controllers/LocationControllerTest.java:482-511`). There is no case for the repository reporting that no row was removed, which is why the `500` has gone unnoticed.

## Acceptance

- A request naming an entity that is not at the given location is answered `404` with the standard `ErrorResponse` body, matching the sibling endpoint.
- `docs/openapi/viron-api.json` describes that answer.
- The case is covered by a `LocationControllerTest` case.

dmccoystephenson:
## Summary

- `PUT /api/v1/locations/{locationId}/entity/{entityId}/move` refused a move into an occupied location by reading occupancy and then writing on the strength of that read, with no boundary between the two. Two moves into the same empty location both read it empty and both committed, so the collision the endpoint's 409 exists to prevent happened anyway (#203).
- The target location's own row is now locked before its occupancy is read, and the move runs inside a transaction so the lock is held until the write is committed. A second move into the same location waits, then reads the placement the first one left behind and is answered with the 409 it would have received had the two requests arrived in sequence.
- The location row is what gets locked, rather than whatever occupies it, because an empty location has no occupancy rows to lock and emptiness is exactly the state the check depends on. Nothing in the schema settles this the way the primary key on `entity_location.entity_id` settles the placement race in #200: a location is permitted to hold several entities, and `addEntityToLocation` places one without consulting occupancy at all. The question of whether one-entity-per-location should become a real invariant — option 1 of #203 — is left untouched by this change, which only makes the rule this one endpoint already applies hold under concurrency.
- The entity's own placement is locked before the target location, because that is the order `EnvironmentController.deleteEnvironment` acquires the same two locks in — it clears `entity_location` before deleting the locations — and the reverse order would let a move and a cascade delete wait on each other in a cycle. Locking the placement first also keeps the position the grid and adjacency checks are made against from moving underneath them.
- `DELETE /api/v1/locations/{locationId}/entity/{entityId}` answered 500 when the entity named was not placed at the location named, because no row was affected and the failure was reported as a `ServiceException`. It now answers 404, matching the sibling `DELETE /api/v1/locations/entity/{entityId}`, which already handled the equivalent case that way (#210). Its check runs under the same placement lock and transaction, so two removals of the same placement at once do not reintroduce the 500 through the back door.
- `DbInteractions.lock` was added for the locking statements. The existing `queryOne` catches a `SQLException`, logs it, and returns an empty `Optional`, which would have made a lock timeout or a broken deadlock indistinguishable from a row that does not exist — and so would have answered 404 for a location that plainly exists. The new method reports the failure as `CannotAcquireLockException` instead, in the manner `updateReportingDuplicateKey` already reports a duplicate key.
- The OpenAPI spec and `docs/MVP.md` are updated for both endpoints, and the Python client's message for the delete's 404 no longer names only the location, since that answer now covers the placement too.

## Left for a follow-up

Nothing bounds how long a request waits for one of the new locks: the pool size is HikariCP's default and no `lock_timeout` is set. That is a deployment-configuration decision that interacts with the cascade delete's statement-per-row loop, so it is filed as #212 rather than settled here.

## Test plan

- [x] `mvn test -B` — 441 tests, 0 failures (`./mvnw` is not usable in this environment; the system Maven was used, and CI runs `./mvnw test -B` against the same suite).
- [x] `python3 -m pytest` — 108 passed.
- [x] `ConcurrentMoveTest` was confirmed to catch the defect rather than merely to pass: it fails both with `FOR UPDATE` removed from the locking statement and with the lock moved to after the occupancy read, and passes only with the lock in place.
- [x] `TransactionBoundaryWiringTest` covers that both endpoints' `@Transactional` is genuinely applied at runtime, since the locks are worthless without it.
- [x] `DbInteractionsTest` covers that a locking statement reports whether it matched a row and reports a failed statement rather than returning false.
- [x] New `LocationControllerTest` cases cover the lock ordering, a target deleted between the two reads, a placement that cannot be read back after being locked, and both shapes of the delete's new 404.

Closes #203
Closes #210

### DIFF
--- a/docs/MVP.md
+++ b/docs/MVP.md
@@ -51,8 +51,8 @@ Deliver a working, tested API that supports:
 - [x] `GET /api/v1/locations/{locationId}/occupied` – Report whether a location holds any entities.
 - [x] `GET /api/v1/locations/{locationId}/neighbors` – Retrieve the locations adjacent to a location within its grid.
 - [x] `PUT /api/v1/locations/{locationId}/entity/{entityId}` – Place an unplaced entity at a location (no-op when the entity is already there, conflict when it is placed elsewhere).
-- [x] `PUT /api/v1/locations/{locationId}/entity/{entityId}/move` – Move a placed entity to an adjacent, unoccupied location in the same grid.
-- [x] `DELETE /api/v1/locations/{locationId}/entity/{entityId}` – Remove an entity from a specific location.
+- [x] `PUT /api/v1/locations/{locationId}/entity/{entityId}/move` – Move a placed entity to an adjacent, unoccupied location in the same grid (concurrent moves into the same location are resolved in sequence, one of them taking effect and the rest reported as conflicts).
+- [x] `DELETE /api/v1/locations/{locationId}/entity/{entityId}` – Remove an entity from a specific location (not found when the entity is not placed there).
 - [x] `DELETE /api/v1/locations/entity/{entityId}` – Remove an entity from its current location.
 
 ---
--- a/docs/openapi/viron-api.json
+++ b/docs/openapi/viron-api.json
@@ -393,6 +393,7 @@
       },
       "delete": {
         "summary": "Remove entity from location",
+        "description": "Removes the entity placed at the location. An entity that is not placed there, whether it is placed elsewhere or not placed at all, is reported as not found.",
         "parameters": [
           { "$ref": "#/components/parameters/LocationIdParam" },
           { "$ref": "#/components/parameters/EntityIdParam" }
@@ -400,7 +401,7 @@
         "responses": {
           "204": { "description": "Entity removed" },
           "404": {
-            "description": "Location not found with that id",
+            "description": "No location with that id, or the entity is not placed at that location",
             "content": {
               "application/json": { "schema": { "$ref": "#/components/schemas/ErrorResponse" } }
             }
@@ -411,6 +412,7 @@
     "/api/v1/locations/{locationId}/entity/{entityId}/move": {
       "put": {
         "summary": "Move an entity to an adjacent, unoccupied location in the same grid",
+        "description": "Moves a placed entity to an adjacent, unoccupied location in the same grid. The target is held against other moves for the duration of the request, so two moves into the same empty location are resolved in sequence: one takes effect and the other is answered as a conflict.",
         "parameters": [
           { "$ref": "#/components/parameters/LocationIdParam" },
           { "$ref": "#/components/parameters/EntityIdParam" }
@@ -430,7 +432,7 @@
             }
           },
           "409": {
-            "description": "Target location is already occupied",
+            "description": "Target location is already occupied, including when a concurrent move reached it first",
             "content": {
               "application/json": { "schema": { "$ref": "#/components/schemas/ErrorResponse" } }
             }
--- a/src/main/java/preponderous/viron/controllers/LocationController.java
+++ b/src/main/java/preponderous/viron/controllers/LocationController.java
@@ -129,16 +129,25 @@ public class LocationController {
      * it is placed elsewhere or not placed at all — is a request naming something that does not
      * exist, answered as such rather than as a server fault (#210); the sibling
      * {@link #removeEntityFromCurrentLocation(int)} already answers the equivalent case the same way.
+     *
+     * <p>The placement is locked before it is read, and the read and the delete run in one
+     * transaction, so that two removals of the same placement at once are resolved in sequence.
+     * Without the lock the second of them would find the row already gone by the time it wrote,
+     * and be answered with the server fault this endpoint has just stopped reporting.
      */
     @DeleteMapping("/{locationId}/entity/{entityId}")
     @ResponseStatus(HttpStatus.NO_CONTENT)
+    @Transactional
     public void removeEntityFromLocation(@PathVariable("entityId") @Min(1) int entityId, @PathVariable("locationId") @Min(1) int locationId) {
         if (locationRepository.findById(locationId).isEmpty()) {
             throw new NotFoundException("Location not found with id: " + locationId);
         }
+        if (!locationRepository.lockPlacementOfEntity(entityId)) {
+            throw entityNotAtLocation(entityId, locationId);
+        }
         Optional<Location> placement = locationRepository.findByEntityId(entityId);
         if (placement.isEmpty() || placement.get().getLocationId() != locationId) {
-            throw new NotFoundException("Entity " + entityId + " is not at location " + locationId);
+            throw entityNotAtLocation(entityId, locationId);
         }
         if (!locationRepository.removeEntityFromLocation(entityId, locationId)) {
             throw new ServiceException("Failed to remove entity " + entityId + " from location " + locationId);
@@ -241,6 +250,10 @@ public class LocationController {
 
     private static NotFoundException entityNotPlaced(int entityId) {
         return new NotFoundException("Entity " + entityId + " is not placed at any location");
+    }
+
+    private static NotFoundException entityNotAtLocation(int entityId, int locationId) {
+        return new NotFoundException("Entity " + entityId + " is not at location " + locationId);
     }
 
     /** True if {@code a} and {@code b} are within one grid cell of each other (Chebyshev distance 1), including diagonals. */
--- a/src/main/java/preponderous/viron/controllers/LocationController.java
+++ b/src/main/java/preponderous/viron/controllers/LocationController.java
@@ -200,14 +200,22 @@ public class LocationController {
      * an invariant of the data. The target's row is therefore locked before its occupancy is read,
      * and the lock is held to the end of the transaction, so a second move into the same location
      * waits and then reads the placement the first one committed.
+     *
+     * <p>The entity's own placement is locked first, before anything is read, for two reasons: it
+     * keeps the position the checks below are made against from moving underneath them, and it is
+     * the order {@code deleteEnvironment} takes the same two locks in, so a move and a cascade
+     * delete cannot end up waiting on each other in a cycle.
      */
     @PutMapping("/{locationId}/entity/{entityId}/move")
     @ResponseStatus(HttpStatus.NO_CONTENT)
     @Transactional
     public void moveEntityToLocation(@PathVariable("entityId") @Min(1) int entityId,
                                      @PathVariable("locationId") @Min(1) int locationId) {
+        if (!locationRepository.lockPlacementOfEntity(entityId)) {
+            throw entityNotPlaced(entityId);
+        }
         Location current = locationRepository.findByEntityId(entityId)
-                .orElseThrow(() -> new NotFoundException("Entity " + entityId + " is not placed at any location"));
+                .orElseThrow(() -> entityNotPlaced(entityId));
         Location target = locationRepository.findById(locationId)
                 .orElseThrow(() -> new NotFoundException("Location not found with id: " + locationId));
         Optional<Integer> currentGrid = locationRepository.getGridIdOfLocation(current.getLocationId());
@@ -229,6 +237,10 @@ public class LocationController {
         if (!locationRepository.moveEntityToLocation(entityId, locationId)) {
             throw new ServiceException("Failed to move entity " + entityId + " to location " + locationId);
         }
+    }
+
+    private static NotFoundException entityNotPlaced(int entityId) {
+        return new NotFoundException("Entity " + entityId + " is not placed at any location");
     }
 
     /** True if {@code a} and {@code b} are within one grid cell of each other (Chebyshev distance 1), including diagonals. */
--- a/src/main/java/preponderous/viron/controllers/LocationController.java
+++ b/src/main/java/preponderous/viron/controllers/LocationController.java
@@ -4,6 +4,7 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.dao.DuplicateKeyException;
 import org.springframework.http.HttpStatus;
+import org.springframework.transaction.annotation.Transactional;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.*;
 import preponderous.viron.dto.LocationDto;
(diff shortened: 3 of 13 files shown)


dmccoystephenson:
## Self-review

The diff was reviewed against this repository's conventions, and the branch was re-run in full afterwards: `mvn test -B` gives 441 tests with 0 failures, and `python3 -m pytest` gives 108 passed. Four findings were raised; three have been fixed on this branch and one has been filed as a follow-up.

The new race test was checked for teeth rather than assumed to have them. Two mutations were tried: removing `FOR UPDATE` from the locking statement, and moving the lock to after the occupancy read. `ConcurrentMoveTest` fails under both and passes only with the lock in place, so it is a regression test rather than a test that happens to pass.

### Fixed in c846d38

- `src/main/java/preponderous/viron/repositories/LocationRepositoryImpl.java:114` — both lock helpers derived their boolean from `DbInteractions.queryOne`, which catches a `SQLException`, logs it, and returns an empty `Optional`. A lock timeout, a deadlock the database had broken, or a serialization failure was therefore indistinguishable from a row that does not exist, and the controller answered `404 "Location not found with id: N"` for a location that plainly exists — telling the client a resource is gone when the truth was that it was momentarily unavailable. Precisely the contention this PR introduces waiting for is what would have triggered it. A `DbInteractions.lock` was added that reports the failure as `CannotAcquireLockException` instead, in the manner `updateReportingDuplicateKey` already reports a duplicate key, and both helpers now go through it.
- `src/main/java/preponderous/viron/controllers/LocationController.java:139` — the new placement check in `removeEntityFromLocation` was itself a check-then-act with nothing held between the two steps. Two removals of the same placement at once both passed the check; the second matched no rows and was answered with the `500` that #210 set out to remove. The placement is now locked first, inside a transaction, exactly as the move does it, and `TransactionBoundaryWiringTest` covers that the boundary is genuinely applied.
- `src/test/java/preponderous/viron/controllers/LocationControllerTest.java:690` — `moveEntityToLocation_EntityNotPlaced` stubbed only `findByEntityId`, a stub that had become dead: the unstubbed `lockPlacementOfEntity` returned `false` and the controller threw before `findByEntityId` was reached, so the test passed for a different reason than it documented and the `orElseThrow` fallback had no coverage at all. The test now stubs the lock explicitly and asserts that nothing further is read, and a second case covers the fallback.

### Fixed in dfc9c0c, before this review

- The target location was originally locked without the entity's placement being locked first, which put the move's two locks in the opposite order to `EnvironmentController.deleteEnvironment` — it clears `entity_location` before deleting the locations. A move and a cascade delete touching the same rows could then have waited on each other in a cycle. The placement is now locked first, and the ordering is asserted in `moveEntityToLocation_LocksTheTargetBeforeReadingItsOccupancy`. `EntityRepositoryImpl.deleteById` was checked as well and takes the same order.

### Filed rather than fixed

- `src/main/java/preponderous/viron/config/DataSourceConfig.java:37` — nothing bounds how long a request waits for one of the new locks. `maximumPoolSize` is left at HikariCP's default of 10, and neither the properties nor the JDBC URL sets `lock_timeout` or `statement_timeout`, whose Postgres defaults are to wait indefinitely. A waiter holds a pooled connection for the whole wait, so sufficient contention drains the pool and unrelated endpoints begin failing. This is a deployment-configuration decision, and a blanket `statement_timeout` would interact with the cascade delete's statement-per-row loop, so it is filed as #212 rather than settled here.

### Checked and found sound

- Both new statements are parameterized, and no JDBC resource is opened outside `DbInteractions`.
- `CannotAcquireLockException` is a Spring `DataAccessException` and reaches `GlobalExceptionHandler`'s catch-all as a 500; whether it deserves a 409 or 503 of its own is part of #212 rather than of this change.
- The fix depends on the default `READ COMMITTED` isolation, under which the waiter's occupancy read takes a fresh snapshot and sees the winner's committed placement. No isolation level is set anywhere in the service, so that default holds.
- No endpoint's path, request shape, or response shape changed, so `OpenApiSpecDriftTest` remains green; the spec and `docs/MVP.md` were updated for the descriptions and the delete's new 404.
- The Python client needed no contract change, since it already treated this delete's 404 as an error; only its message was corrected, as that message now covers the placement as well as the location.

dmccoystephenson:
## Review of the whole branch (second pass)

The full diff was re-read against this repository's conventions in a later session, after the previous one was interrupted mid-cycle. The rubric is scored below against the diff and against command output rather than against judgement, and two findings were raised that the first pass did not: one has been fixed on this branch and one has been filed.

The external anchor is genuinely green this time. The run on `c846d38` was recorded as a failure with all three of its jobs still queued after thirty minutes — an abandoned run rather than a verdict on the diff — and rerunning it was not available to this session, so the fix below was pushed and CI ran clean on `8e1f4c6`: `build`, `python-client (3.8)`, and `python-client (3.12)` all pass. Locally, `mvn test -B` gives 443 tests with 0 failures and `pytest` gives 108 passed on both interpreters present here.

### Findings

**Fixed in 8e1f4c6**

`src/main/java/preponderous/viron/controllers/LocationController.java:157` — `removeEntityFromCurrentLocation` kept exactly the shape #210 removes from its sibling. It read the placement, found it present, and then deleted on the strength of that read, with no lock and no transaction around the pair. Two removals of the same placement therefore both passed the check; the one that wrote second matched no rows, `removeEntityFromCurrentLocation` returned `false`, and the request was answered with the same `ServiceException`-backed 500 that this PR exists to stop reporting for a removal that merely lost a race. The endpoint mattered here in particular because the javadoc this PR adds to the other removal cites it as the sibling that "already answers the equivalent case the same way" — true of the 404 for an unplaced entity, but not of the concurrent case. It now takes `lockPlacementOfEntity` inside a `@Transactional` boundary, in the same order and the same manner as its sibling. The preceding `findByEntityId` was dropped rather than kept, because a locking statement that matches no row is precisely the unplaced entity the 404 reports, so a second read would only widen the window it was there to close. No status code, message, or body of that endpoint changes, so the OpenAPI spec, `docs/MVP.md`, and the Python client need no amendment for it.

The fix was checked for teeth rather than assumed to have them: with the controller change stashed, three of the four `removeEntityFromCurrentLocation` cases fail, and all four pass with it restored.

**Filed rather than fixed**

`src/main/java/preponderous/viron/controllers/LocationController.java:85` — the lock this PR adds excludes other moves and nothing else. `addEntityToLocation` neither consults the target's occupancy nor takes its lock, so a placement and a move aimed at the same empty location can still both succeed and leave two entities there. The PR body already says the invariant question is left open as option 1 of #203, but #203 is closed by this merge, which would have taken the open question with it. It is now #213, with the three concurrent pairs tabulated and the two coherent answers set out. This is not a regression — the state has always been one the schema permits — but the move's documented 409 reads as a guarantee about the location when it is only a guarantee about other moves.

### Rubric

- **Scope:** PASS — all thirteen modified files belong to the move race (#203), the removal's status code (#210), or the removal-endpoint consistency finding above; no unrelated formatting or renames appear in `git diff origin/main...HEAD`.
- **Tests-new:** PASS — `lockPlacementOfEntity`, `lockLocation`, and `DbInteractions.lock` each have direct cases in `LocationRepositoryImplTest` and `DbInteractionsTest`; both new private helpers are exercised through the controller cases.
- **Tests-fix:** PASS — established empirically, not by reasoning. `ConcurrentMoveTest` fails with `FOR UPDATE` removed and with the lock moved after the occupancy read; the `removeEntityFromCurrentLocation` cases fail with the controller change stashed.
- **Sibling structure:** PASS — `ConcurrentMoveTest` follows `ConcurrentPlacementTest`'s H2-per-test shape, and the new repository methods sit with their neighbours in `LocationRepositoryImpl`.
- **Sibling renames:** PASS — no identifier in a parallel series was renamed; the two new `NotFoundException` factories are additions.
- **Docs:** PASS — the spec and `docs/MVP.md` carry both changed endpoints; `docs/PLANNING.md`, `tickets.md`, `README.md`, and the Postman collection describe these endpoints only at a level that remains accurate, and were re-checked line by line.
- **Issue resolution:** PASS — #203's acceptance is met by `ConcurrentMoveTest`, and #210's by the two new 404 cases; neither is claimed while partially resolved.
- **CI:** PASS — green on `8e1f4c6`, all three jobs.
- **DTO boundary:** PASS — no controller method in the diff returns `Entity`, `Environment`, `Grid`, or `Location`; the two changed endpoints return `204 No Content`.
- **Spec alignment:** PASS — no path, verb, parameter, or body shape changed. The spec edits are descriptions and the delete's 404 wording, and `OpenApiSpecDriftTest` is green.
- **Java/Python parallelism:** PASS — the only client-visible change is the delete's 404 message, which `locationService.py` and its test carry.
- **Override correctness:** PASS — every `@Override` in the diff (`LocationRepositoryImpl`'s two new methods, `LockRacer.lockLocation`) matches a real interface or superclass method, which the compile step confirms.
- **Scope ceiling:** PASS — 147 net non-test lines across seven non-test files (`git diff --stat origin/main...HEAD -- src/main/java src/main/python docs`), well inside the ceiling; the remaining 491 of the diff's 645 added lines are tests.

### Observations outside the diff

- `tickets.md:165` still lists "Adding entity to new location removes from old location" as acceptance criteria for the location endpoints. `addEntityToLocation` has since been changed to answer a conflict for an entity placed elsewhere, so that line describes behaviour the service no longer has. It is a record of a completed ticket rather than live documentation, and it predates this branch, so it is left alone here.
- `CannotAcquireLockException` reaches `GlobalExceptionHandler`'s catch-all as a 500. Whether contention deserves a 503 or a 409 of its own belongs with #212, where the timeouts that would produce it are being decided.

---

dmccoystephenson:
## Merge readiness — held for the codeowner

Everything this PR needs has been done, and it is being left open on one gate rather than on any remaining work.

**Green.** CI passes on `8e1f4c6` — `build`, `python-client (3.8)`, and `python-client (3.12)`. Locally, `mvn test -B` gives 443 tests with 0 failures and `pytest` gives 108 passed. The branch is `MERGEABLE` and `CLEAN` against `main`. The regression gate is met for both bug fixes, and both regression tests were confirmed to fail without their fix rather than merely to pass with it. The documentation sources of truth were re-checked line by line against the implementation.

**The gate.** One modified path is on this repository's do-not-auto-merge list, so autonomous merge is withheld:

- `docs/openapi/viron-api.json` — the API contract.

What the spec change amounts to is worth stating plainly, since the hold is on the path rather than on the size of the edit. Four lines move: two `description` fields are added, and two response descriptions are reworded. No path, verb, parameter, request body, response schema, or status code is added or removed, and `OpenApiSpecDriftTest` is green. What does change behind it is observable, however, and is the reason a human should look: `DELETE /api/v1/locations/{locationId}/entity/{entityId}` answers `404` where it previously answered `500` for an entity that is not placed at the location named (#210).

No other protected path is touched — no workflow, no `pom.xml`, no `requirements.txt`, no deployment or database bootstrap file — and no file in the diff deletes more than six lines.

**What is being asked.** A codeowner's approval of the contract wording and of the `500` → `404` change on that endpoint. With that, the PR can be squashed as it stands; nothing further is outstanding on the branch.

---

